PR #8, opened 2016-02-21
Update build scripts for Git, upgrade to Python 2.7

### LINKED ISSUE #5: Upgrade to Python 2.7
I'd like to see the first Git release of EventGhost, v0.5.0, include Python 2.7. When #2 is complete, we'll begin investigating this possibility.

topic2k:
The extension needs to be rebuild for py27:

- lib27\site-packages\_dxJoysticks.pyd
- lib27\site-packages\cFunctions.pyd
- plugins\System\VistaVolEvents.pyd

Resolves #5.


blackwind:
I haven't put together a build environment yet, but here are my initial thoughts on your changes:

`[various files]`
* Why is self.enabled being reversed in a bunch of places?

`tools\builder\CreateLibrary.py`
* [line 150] Manifest is required for a project Pako and I have been working on and needs to be restored.
* [line 214] Is _ssl.pyd being built now? If not, we need to find a way to build it, and if so, can we remove it from neededDlls here?

`tools\builder\CreatePyExe.py`
* [lines 55, 71, 78] Restore manifest.

`tools\builder\Tasks.py`
* Since most of our change history wasn't documented, let's start fresh here with a new file in a new format.
* Change format to "YYYY-MM-DD (0.5.0):\n* Commit message 1 (author)\n* Commit message 2 (author)\n\n". Filename can stay the same.
* [line 62] Changelog task should be able to handle a missing or empty file.

By and large, the rest looks good to my eye, but we'll see how it holds up when I begin testing in a different environment.

blackwind:
Well, I'm right on the cusp here, but can't quite make it over the finish line. My last remaining dependency, `pywin32`, fails to install with the following error:

    close failed in file object destructor:
    sys.excepthook is missing
    lost sys.stderr

Despite that, enough of its files appear to be in place to run `Build.py`, but the build dies when it tries to import the module:

      File "c:\usr\local\python2\Lib\threading.py", line 810, in __bootstrap_inner
        self.run()
      File "c:\usr\local\python2\Lib\threading.py", line 763, in run
        self.__target(*self.__args, **self.__kwargs)
      File "[...]\EventGhost\tools\builder\Gui.py", line 97, in DoMain
        builder.Tasks.Main(self.buildSetup)
      File "[...]\EventGhost\tools\builder\Tasks.py", line 235, in Main
        task.DoTask()
      File "[...]\EventGhost\tools\builder\CreateImports.py", line 293, in DoTask
        for module in GetPackageModules(package):
      File "[...]\EventGhost\tools\builder\CreateImports.py", line 223, in GetPackageModules
        mod = __import__(package)
    ImportError: No module named pywin32

I'm off to bed now, so I'm posting this as both a status update and a call for help. If you have any ideas as to how to get `pywin32` installed properly, I'm all ears; if not, I'll resume Googling in the morning.

topic2k:
:-)  hehe, you go to bed and i have my first pause at work.
well, i had no problem installing pywin32. Are you building from source? I used the installer from SourceForge: [pywin32-220.win32-py2.7.exe](https://sourceforge.net/projects/pywin32/files/pywin32/Build%20220/pywin32-220.win32-py2.7.exe/download)

CarsonF:
If I could just say two things. I think the upgrade to 2.7 should be separate for the modifications needed for git. Not that these couldn't happen one right after the other. But it's hard to see from this diff what is actually changing with the build tool. Smaller specific PRs are always better than bigger ones.

The second is more of a styling point. Many developers feel its a good practice to commit whitespace changes for files/code that aren't actually changing. It makes it harder to read what is actually changing in the code base. 

IF we want to do want to do a styling pass I would suggest that would be separated as well.

blackwind:
> I googled and found this:

Yep, that's probably it. Will confirm in the coming hours.

> If I could just say two things. I think the upgrade to 2.7 should be separate for the modifications needed for git. Not that these couldn't happen one right after the other. But it's hard to see from this diff what is actually changing with the build tool. Smaller specific PRs are always better than bigger ones.

I agree wholeheartedly and was going to say something myself, but I wasn't sure how much of a pain in the ass asking him to redo all the commits would've been. In the coming days, I'll be moving one of my pet projects to GitHub to get a better feel for everything.

> IF we want to do want to do a styling pass I would suggest that would be separated as well.

Agreed. I'm in favor of the styling pass, and will be looking at something like [EditorConfig](http://editorconfig.org/) (unless, again, you have a better suggestion) to enforce it going forward.

topic2k:
> Why is self.enabled being reversed in a bunch of places?

Thats a little logic to decide which tasks could be skipped when building the installer. While trying to understand the code i maybe messed it a little. I've tried to clean it up.


> Manifest is required for a project Pako and I have been working on and needs to be restored.

If it's another project (plugin?), wouldn't in then be better to move it there and not directly to EG?

>  Is _ssl.pyd being built now? If not, we need to find a way to build it, and if so, can we remove it from neededDlls here?

The standard Python 2.7 Stackless install has a `_ssl.pyd` (in dll folder). I removed it from neededDlls and made a test build. The dll will be included in `EG\lib27\`

>  Since most of our change history wasn't documented, let's start fresh here with an empty file in a new format.
    [line 62] Changelog task should be able to handle a missing or empty file.
    [line 123] Extract only the first line (the title/subject) of the commit, but leave the existing code in place (commented out) in case we want to switch back. End-users probably don't need all the gory details of each commit.

Done.


> I think the upgrade to 2.7 should be separate for the modifications needed for git.

Well, i'm not yet that familar with git and don't know how to split those now. Do i need all commit history for that? That would be a problem because in between i deleted my fork and forked it new.
And because we make here a new start, i think that's ok. From what i read from the changes between Python 2.6 and 2.7.
After that's done, i agree to do smaller commits. :)

blackwind:
> If it's another project (plugin?), wouldn't in then be better to move it there and not directly to EG?

To clarify, I used the term "project" because it's an ongoing piece of work we're doing. It's related to the EG core and isn't a plugin. The manifest definitely needs to be restored.

I'll take a look at the rest of this in the coming hours.

CarsonF:
@topic2k How come you deleted your fork? It would really be better if we had the history here. And I think upgrading to 2.7 might be too big of a bit to take all at once. I would rather get the builder running with git. Do a stable release, and then upgrade to 2.7. My other thought is with that release we don't have to include all of the dependencies manually.

CarsonF:
Also EditorConfig is a great start :+1:

blackwind:
I still like including Python 2.7 in our first Git release. There's nothing forcing us to release right away. Once I get this built, I'll be testing it in production for at least a week or two.

That said, I still agree that it would be great if we had [atomic commits](http://www.freshconsulting.com/atomic-commits/) with [well-written commit messages](http://chris.beams.io/posts/git-commit/) and a separate pull request for each of the three parts of this.

blackwind:
So, the good news is that the new commits look good and I was able to get `pywin32` installed. The bad news is, I still can't complete the build.

Have you tested this yet with all the boxes checked? When "Create Imports.py" is checked, I always get the error above (because there _is_ no module named `pywin32`, but rather, a collection of modules starting with `win32*`); when "Create Imports.py" _isn't_ checked, the installer gets built, but if you install it and try to run the program, it fails spectacularly because the imports file wasn't created successfully.

Carson is working on putting together a build environment of his own to confirm that this isn't a case of me doing something wrong (I don't think it is, but it can't hurt to verify), but it doesn't look like he finished before bed. I suspect he'll post a follow-up before I wake up. In the meantime, make sure _all_ steps of the build process are working on your end.

topic2k:
Yes, i successfully build with all steps enabled except the upload and website sync. When i'm home in the evening i will Setup a virtualbox and try it there.

blackwind:
Weird. Even my current EventGhost installation doesn't have a module named `pywin32` (confirmed via PyCrust), so how is it that you have one? I'm very interested to find out what exactly it is I'm doing wrong.

topic2k:
Do you do import `pywin32`? That fails for me too. I think that's right. For me it looks like `win32com`, `win32api`, `win32file` etc. are the real imports.

blackwind:
That's the conclusion I came to as well, but `CreateImports.py` still tries to import a module called "pywin32" (which should never exist anywhere) and dies as a result. And yet, the build completes for you, so it must not try to import it on your end for some reason that we need to figure out.

topic2k:
Hello blackwind, the script can't find the .pth file for pywin32 (`Python27\lib\site-packages\pywin32.pth`).
Mine has the following contents:

    .pth file for the PyWin32 extensions
    win32
    win32\lib
    Pythonwin

Please create it and try again.

blackwind:
No luck. `pywin32.pth` is already in the right place and contains the right contents.

I have a bad feeling the build script doesn't like me using a virtualenv. Will investigate in the coming hours.

blackwind:
Sure enough, virtualenv issue. I've gone ahead and opened #12 for that. In the meantime, I've merged the pull request, upgraded the extensions, and made a few smaller changes. I'm now running the resulting build in production, and so far so good on that front.

Issue #16, opened 2016-02-24
Enable continuous integration


blackwind:
We now have a [Travis CI](https://travis-ci.org/) build status icon in the [README](https://github.com/EventGhost/EventGhost/blob/master/README.md), but currently have no build instructions for it to follow. I'll take a look at setting this up in the coming days.

Issue #15, opened 2016-02-24
Generate changelog from pull requests instead of commits


blackwind:
@topic2k's build script update (#8) contains the ability to generate a changelog from Git commits. I suggested we only include the title/subject of each commit rather than the full body, but I fear this will still produce far too much noise for the end-user.

Consider the following recent commit messages:

* "Added README with project description and slack link"
* "Restored manifest file."
* "Merge pull request #8 from topic2k/develop"
* "Add .gitignore"

Do we really want all that in the changelog? Does the user care about any of these things?

Here's what I'm thinking we should include, ordered by order in the changelog:

* Merged pull requests with the `notice` tag ("Important changes for plugin developers:")
* Merged pull requests with the `enhancement` tag ("Features added:")
    * Except where also tagged with `internal`
* Merged pull requests with the `bug` tag ("Bugs fixed:")
    * Except where also tagged with `internal`

There's a Ruby project called [GitHub Changelog Generator](https://github.com/skywinder/github-changelog-generator) capable of producing this result. I don't think we should use that project directly to generate our changelog, as the last thing we need is a dependency on another language, but I think we should aim to reproduce [its excellent Markdown output format](https://github.com/skywinder/Github-Changelog-Generator/blob/master/CHANGELOG.md).

Added bonuses:

* If this completely eliminates the need for `pygit2` (and, by extension, libgit2), the process of setting up a build environment will be greatly simplified.
* We can publish the Markdown output to our GitHub releases (#14).

We need to resolve this before closing #2.

blackwind:
I've been thinking about the content of the above post, and I've identified a potential stumbling point -- we can compare the date an issue was closed with the date of the last release, but does that tell us the whole story?

Consider the following scenario: We release v0.5.0, then we begin work on v1.0.0. For v1.0.0, hypothetically, we upgrade to Python 3. We've converted the source tree to Python 3, but we're not ready to release it, and suddenly the need arises to put out a patch release, v0.5.1. We checkout the `v0.5.0` tag, commit our bugfix, begin building the v0.5.1 release, and the changelog generator picks up and adds the closed "Upgrade to Python 3" issue. Not good.

I don't think there's any way to reliably generate a changelog from issues. Pull requests, though? Not only can that still work, it's probably what I should've suggested in the first place, as every closed issue listed would've had a duplicate in the list of merged pull requests.

[...]

PR #25, opened 2016-03-03
Add build step to publish update to GitHub

### LINKED ISSUE #14: Add build step to publish update to GitHub
As previously discussed, we should add an option to the build script, NOT enabled by default, to publish our update to GitHub. Steps taken would include:

* Generate changelog.
* Push changelog to remote.
* Add GitHub release in format "v0.5.0".
* Upload installer to release.
* Publish changelog to release.

We need to resolve this before closing #2.

topic2k:
- create changelog from GitHub (makes pygit2 obsolete)
- commit the changelog
- change the reference to the commit
- create release
- upload the setup file to the release

The builder gui is extended with controls for the needed github info: token, user, repo, branch. These infos will be saved in the builder config file.

Resolves #14.


blackwind:
I won't have time to fully review this until late tonight, but one thing I wanted to point out before leaving:

`build\builder\Tasks.py`

* [lines 38, 40] This is `pygit2` code I recently added that will need to be updated. All `len(list(repo.walk(repo.head.target)))` does is count the total number of commits.

topic2k:
Didn't saw that, sorry. I can count the commits on GitHub, but it's a little bit slow. Should i make another PR or add it to this? (But not today anymore, going to bed.) Tomorrow i will investigate if there is a faster method than counting.

blackwind:
No need to apologize. I've been going hard these past few days while I still had oodles of time on my hands. If anyone has actually stayed on top of all my changes, I'd be stunned.

In general, always add to the existing pull request if it's the same project and it hasn't already been merged. The good news is, mentally and physically, I'm completely and utterly exhausted, so I can say with confidence that there won't be any merges (or likely any code review at all) tonight. :stuck_out_tongue:

topic2k:
Found a way do get the revision (commit) count from via github api. I removed the pygit2 dependency and made successfull a complete build. (All Tasks up to and including releasing to github.)

blackwind:
Good stuff. I'll make every effort to get this merged by the end of the day.

Regarding the pre-release checkbox idea, I like it, but let's continue that discussion over in #15.

blackwind:
`[various files]`

* Let's read the user's GitHub token from their `.gitconfig` using [this standard](https://github.com/blog/180-local-github-config). No need to add code to write to `.gitconfig` -- just exit and tell the user to run `git config --global github.token [token]` if you can't find it.
* Let's remove the user field to prevent confusion and instead store the repo in format "EventGhost/EventGhost". If I saw the user field without looking at the code, I'd probably put "blackwind" in there instead of "EventGhost" and end up releasing to my fork by mistake.
* The `while nextPage:` blocks could be condensed into a function so there isn't so much code duplication.

`build/builder/CheckDependencies.py`

* I was going to suggest that you require `agithub` v2.0, then delete `build/builder/github.py`, but apparently v2.0 isn't on PyPI yet. I'll make this change myself when that changes.

`build/builder/CreateGitHubRelease.py`

* [line 1-15] Replace file header with our new file header found in all .py files.
* [line 72] Change to `print "getting release info"`.
* [line 74] Change to `release = None`, new line, `upload = True`. We'll be getting rid of `add2release` entirely further down.
* [line 76] This should probably be an error with a `return`. Even now, we have a `v0.4.1.r1722` release tag, so there's no reason this call should ever fail.
* [line 78] Consider the v0.5.1 release scenario I outlined in #15, and hypothetically, pretend we released a v1.0.0-beta1. This code will look at v1.0.0-beta1, decide it's safe to continue, and potentially do something we don't want. You'll probably have to loop through all releases. If you find a matching release already exists, save it under the `release` variable.
* [line 84] Change to `wx.YES|wx.NO`. "Cancel" suggests to the user that the process will be cancelled entirely, which it isn't.
* [line 85] For code clarity, change to `upload = (True if dlg.ShowModal() == wx.ID_YES else False)`.
* [line 117-155] Before posting, we should check to see if the local changelog is the same as the remote changelog, which could happen if we previously failed somewhere during publishing. If so, skip these steps.
* [line 138] Change to `print "creating commit for changelog update"`.
* [line 139] Change to `body = {'message': "Add changelog for v{0}".format(appVer),`.
* [line 169] Creating a release shouldn't depend on how the user responds to the upload dialog box. We can't create a release if it's already there. Instead, we should do `if not release:` here, then add an accompanying `else:` block to overwrite the changelog on the existing release if you found one above.
* [line 184] Currently, the installer is uploaded even if the user says cancel/no. Do `if upload:` here.

`build/builder/Gui.py`

* `appRevision` isn't populated unless the refresh button is pressed, and `appVersion` and `appRevision` need to be updated before build if repo or branch are changed, so we'll have to come up with another approach here.
* Remove User and Token fields.

`build/builder/UpdateChangeLog.py`

* [line 71] Change to `pass`.
* [line 72] Change to `newline = "- {0} ({1})\n".format(msg, author)`.
* [line 72-73] Move these into the `try:` block. We don't want any empty lines in the changelog.

`build/builder/Utils.py`

* [line 78] Change to `parts = GetLastReleaseOrTagName(buildSetup).split('.')[:3]`.
* [line 80-84] Change to `while len(parts) < 3:`, `parts.append("0")`.

topic2k:
> I was going to suggest that you require agithub v2.0, then delete build/builder/github.py, but apparently v2.0 isn't on PyPI yet. I'll make this change myself when that changes.

No need for a dependency to agithub. The needed script i included in `builder/github.py`.


> appRevision isn't populated unless the refresh button is pressed, and appVersion and appRevision need to be updated before build if repo or branch are changed, so we'll have to come up with another approach here.

That's only on first start when there is no saved user/token. But with reading from `.gitconfig` this will be obsolete.


The next days i don't have much time to work on it. Next weekend i will continue work on your comments. (Just to let you now about delaying.)

blackwind:
> No need for a dependency to agithub. The needed script i included in `builder/github.py`.

I still prefer not to include any non-EventGhost code in the tree that we don't have to. The user needs to `pip install` a boatload of other things to create their build environment anyway, so one more won't hurt anyone -- particularly when we use `requirements.txt` to automate it, which is something I'll be looking at soon in preparation for #16.

> That's only on first start when there is no saved user/token. But with reading from `.gitconfig` this will be obsolete.

Gotcha.

> The next days i don't have much time to work on it. Next weekend i will continue work on your comments. (Just to let you now about delaying.)

No worries. Thanks for the heads-up. Frankly, I could probably use a few days off myself.

topic2k:
> Let's remove the user field to prevent confusion and instead store the repo in format "EventGhost/EventGhost". If I saw the user field without looking at the code, I'd probably put "blackwind" in there instead of "EventGhost" and end up releasing to my fork by mistake.

I can reduce it to one textfield, but i would leave the username changeable because not everyone has the rights to release to "EventGhost/EventGhost". Or did i misunderstood you?

blackwind:
Reduce it to one field is exactly what I meant. Store it in format "EventGhost/EventGhost", and also show it to the user in format "EventGhost/EventGhost". If user johnsmith forks us and changes the field to "johnsmith/EventGhost", it'll release to his fork as expected.

topic2k:
> [line 84] Change to wx.YES|wx.NO. "Cancel" suggests to the user that the process will be cancelled entirely, which it isn't.

What about Yes/Cancel? Cancel will cancel the task. IMO it doesn't make sense to continue if there is a release and the user don't want to upload the installer.

blackwind:
Good point. Stick with Yes/No, but change the wording of the dialog box to "Found an existing GitHub release matching 'v{0}'. Overwrite?", and match the program flow to that wording. If the user says no, `return`; if the user says yes, overwrite the installer and changelog for the existing release. You might need to delete the existing installer in the release first so you don't end up with a duplicate.

blackwind:
Ah, looks like you snuck a commit in here. I don't get an email notification when that happens, so be sure to post a follow-up comment in the future.

`build/builder/CreateGitHubRelease.py`

* [line 81, 100, 204] Remove these lines. No need for the `upload` variable anymore now that you're `return`ing on "No".
* [line 205] Change to `else:`.

`build/builder/Gui.py`

* [line 109-110, 119-121, 124, 198] Remove these lines. Revision was only restored for API compatibility just in case any plugin authors are checking `eg.revision` to determine whether or not a certain feature or bugfix is available. Moving forward, revision is no longer part of our versioning scheme, so it's something we don't want to draw attention to. Come v1.0, it'll be removed entirely.

`build/builder/Utils.py`

* [line 240] Change to `if "token" not in gitcfg or "user" not in gitcfg:` and modify the comment on the line above. User could have a `[github]` section without both necessary values.

`build/builder/__init__.py`

* [line 66] Change to `"To create a token, go to: https://github.com/settings/tokens"`.

topic2k:
Ok, now with a comment: Crossing fingers that the PR will now be accepted.  :smile:

blackwind:
Perfect.

topic2k:
Found a little error in 'builder\Utils.py': Lines 259-273 should be unindented by one. Otherwise only Repositorys with the name "EventGhost" will be listet in the repo choice button. But this will only affect someone who renames his repository.

blackwind:
Yep, that was an intentional change from me. The odds of someone wanting to create releases from their fork are almost zero in the first place, but the odds of someone wanting to create releases from a renamed fork? I just can't picture that ever happening. In the unlikely event that it does, the user in question can unindent. As for those of us who'll actually be creating releases, we get improved startup times and less riffraff this way. When the plugin repository hits and there are 100 plugin repos under the EventGhost account, it's going to make a HUGE difference.

PR #28, opened 2016-03-19
Upgrade to the latest wxPython

### LINKED ISSUE #9: Upgrade to the latest wxPython
If a version of wxPython > v2.8 is the default version, the main frame of EG is shown, but then the app hangs.

topic2k:
This will make EG start/run with wxPython 3.0.2 (resolves #9).  (Only did a quick check of the plugins in this repo.)

**_For (plugin) developers:_**
Take a look at the [changes since wxPython 2.8](http://docs.wxwidgets.org/stable/overview_changes_since28.html). There are not too much.
What is not described there and we found out:
- If you use a `GridBagSizer` with growable rows/columns, be sure to set the growable row/column after you added the corresponding row/column. Example:
```
sizer = wx.GridBagSizer()
sizer.AddMany([
    (self.slider, (0, 0), (1, 3), wx.EXPAND),
    (st1, (1, 0), (1, 1), wx.ALIGN_LEFT),
    (self.valueLabelCtrl, (1, 1), (1, 1), wx.ALIGN_CENTER_HORIZONTAL),
    (st2, (1, 2), (1, 1), wx.ALIGN_RIGHT),
    ])
sizer.AddGrowableCol(1, 1)
```


blackwind:
Here are a few errors I found:

```
C:\Program Files (x86)\EventGhost\eg\Classes\MainFrame\TreeCtrl.py:710: wxPyDeprecationWarning: Call to deprecated item. 
  while wx.GetMouseState().LeftDown():
```

```
Traceback (most recent call last) (1816):
  File "C:\Program Files (x86)\EventGhost\eg\Classes\TaskletDialog.py", line 61, in ProcessingTask
  File "C:\Program Files (x86)\EventGhost\eg\Classes\ConfigDialog.py", line 166, in Configure
    treeItem.Configure(*args)
  File "C:\Program Files (x86)\EventGhost\eg\Classes\ActionItem.py", line 103, in Configure
    return self.executable.Configure(*args)
  File "C:\Program Files (x86)\EventGhost\plugins\System\Execute.py", line 270, in Configure
    lowerSizer.AddGrowableCol(1)
  File "wx\_core.pyc", line 15368, in AddGrowableCol
PyAssertionError: C++ assertion "!m_cols || idx < (size_t)m_cols" failed at ..\..\src\common\sizer.cpp(1980) in wxFlexGridSizer::AddGrowableCol(): invalid column index
```

The `AddGrowableCol` error shows up in a bunch of places. Search all files in the tree for "AddGrowableCol". Each one will need to be fixed.

Finally, here's my only note on the code thus far, which isn't even your fault:

`build\builder\CreateImports.py`

* [line 39] Insert `"wx.build.config",` here so build doesn't fail in a virtualenv.

topic2k:
Ok, i hope i now fetched all `AddGrowable...`.
But now the PR conflicts and i don't know how to resolve it (the easy way).

blackwind:
Should be as simple as:

    git checkout wx_302
    git rebase upstream master
    git push --force

Testing now. So far so good, but I'm gonna give it a few days before merging.

blackwind:
One more: Exiting EventGhost with File > Exit doesn't actually exit, but instead sucks CPU indefinitely until the process is terminated by force. It appears to be getting caught in an infinite loop at line 93 in `plugins\Mouse\__init__.py`.

topic2k:
> One more: Exiting EventGhost with File > Exit doesn't actually exit, but instead sucks CPU indefinitely until the process is terminated by force. It appears to be getting caught in an infinite loop at line 93 in plugins\Mouse\__init__.py.

I can't reproduce that. I've added a `print "looping"` after the `while True:` and it's always stopping fine.

blackwind:
Interesting, must be something in my config. I'll investigate further.

blackwind:
Got it: It's the [Snarl plugin](https://sourceforge.net/p/eventghost/code/HEAD/tree/branches/pako2/not-included-plugins/Snarl/). I can reproduce in an empty config with only that plugin loaded. Could you pastebin me a fixed version?

topic2k:
Replaced with PR #35

PR #31, opened 2016-03-20
PIL changes: replace tostring with tobytes.

topic2k:


blackwind:
Had to fix this in another file as well, so I made the commit myself. Thanks for bringing it to my attention.

PR #34, opened 2016-03-22
Added a new event before leaving mainloop.

topic2k:
Send an event before the mainloop exits, so that plugins can be stopped clean. (e.g. Snarl plugin)


topic2k:
Replaced/integrated into PR #35

PR #37, opened 2016-03-27
Only copy core plugins to `Program Files\EventGhost\plugins`

topic2k:
Only copy core plugins to `Program Files\EventGhost\plugins`. All other plugins to `ProgramData\EventGhost\plugins`.
To mark a plugin as core for the build tool, create a file with the name `core-plugin` inside the plugin folder.

### DIFF
--- a/build/Build.py
+++ b/build/Build.py
@@ -129,9 +129,13 @@ class MyBuilder(builder.Builder):
         files = set(ListDir(srcDir, EXCLUDE_DIRS, fullpath=False))
 
         noincludes = ["."]
-        for f in files:
+        coreplugins = []
+        for f in files.copy():
             if f.endswith("noinclude"):
                 noincludes.append(f.replace("noinclude", ""))
+            elif f.endswith("core-plugin"):
+                coreplugins.append(f.replace("core-plugin", ""))
+                files.remove(f)
 
         installFiles = []
         for f in files:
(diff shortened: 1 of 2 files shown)


blackwind:
What's the rationale for this one? Separating out the core plugins is an important part of the online plugin repository work, but not something I can think of any good reason to do at this stage.

topic2k:
I think that core plugins and optional plugins should be separated, so that it is distinguishable between need and additional plugins. And the `ProgramData\EventGhost\plugins` is normaly writable by the user.

If you plan, not to supply any optional plugin with future releases, then your are right and this patch isn't needed. But i would prefer to keep some additional plugins in the installer, plugins that have no dependency (eg. OSE, OSM, DesktopRemote). [I have in mind to rework the default example config]

blackwind:
Problem is, the user then has 100 plugins lumped in with the ones he installed manually and can't tell at a glance which are his and which are from the installer.

What I'm going to do is accept this, but then comment out most of the new stuff to get us back to the way it was. Then, when we need this code, it'll be there, ready to go.

If I could make a tiny request, in the future, try to avoid including a trailing period in pull request and commit titles. Titles aren't supposed to be sentences.

Pako2:
I think it is right and Bitmonster also intended that way.
However, there is one issue that needs to be solved.
The problem, it is a **language editor** (you have to use the "-translate" commandline option).
Inside the language editor, there are only seen those plugins that are located in the `Program Files\EventGhost\plugins`.
This means that you can not do the translation for those plugins that are located in the `ProgramData\EventGhost\plugins`.

PR #39, opened 2016-03-28
Catch an exception, if `EventGhost.chm` is not writable

topic2k:


blackwind:
I'm not a fan of this one. If anything, we should be moving in the opposite direction -- removing error handling in the build scripts entirely. We don't want builds to complete if they won't contain the necessary files, and a traceback will tell us exactly what went wrong, whereas here, it just says the file couldn't be copied. Why not?

This is especially important for #16, which I'm still (half-heartedly) designing and working on. If we handle errors like this, CI will claim the build completed successfully when it actually didn't, defeating the entire purpose of setting up CI.

PR #45, opened 2016-04-03
Start in build folder when virtualenv activated manually

blackwind:
This is the kind of thing I'd typically commit directly, but I'm opening a pull request to test CI.


blackwind:
Beautiful.

PR #48, opened 2016-04-07
Calculate Python Script indent size from script

blackwind:
My preferred indent size is two spaces. With this patch, the Python Script editor will examine my script and respect my choice rather than foisting its default of four spaces on me.


blackwind:
It appears CI build will fail if you merge before it starts. Good to know.

PR #51, opened 2016-04-21
Restore automatic and manual update checking

topic2k:
- options ->  check on startup
  - help menu -> manually check for update

Currently it uses a GitHub token from my account to get the latest release information. If the PR is accepted, maybe it could be replaced with a token from EventGhost organization? (No [scopes ](https://developer.github.com/v3/oauth/#scopes)needed, just read access for public info)


blackwind:
Good stuff. My only concern about this one is pre-releases. Users running a pre-release should get notifications for all releases, while users running a stable release should only get notifications for stable releases. There should be an additional option (right below `checkUpdateCtrl`), disabled by default, called "Always notify about new pre-releases" so stable users can get pre-release notifications if they want them. This option should be grayed out if the user disables the main check update option.

blackwind:
Oh, and regarding the token, it's not currently possible to create tokens for an organization, but according to GitHub:

> For requests using Basic Authentication or OAuth, you can make up to 5,000 requests per hour. For unauthenticated requests, the rate limit allows you to make up to 60 requests per hour.

Since we're only making one request per day, just do it unauthenticated.

topic2k:
>  My only concern about this one is pre-releases.

Yeah, your right, didn't thought about that one. We should use the GitHub prerelease marker when uploading alpha-, beta-, rc-, etc.-Versions. Then those won't be seen as new(latest) release.

Regardless this i will try to implement the pre-release check.

blackwind:
> We should use the GitHub prerelease marker when uploading alpha-, beta-, rc-, etc.-Versions.

Agreed. Go ahead and include an update for the GitHub build step in this pull request since it's related. You can now easily determine whether or not a version is a pre-release via `if "-" in buildSetup.appVersion` thanks to `builder.Utils.ParseVersion()`, which throws `InvalidVersion` if the given version string isn't formatted to its liking. All valid pre-release version strings contain "-".

topic2k:
Pre-release functionality added to update check and builder.

blackwind:
I prefer to automate things wherever possible so as not to end up with a result we don't want. Offering the user the ability to set the pre-release flag by hand will surely result in someone, at some point, releasing a stable version string as a pre-release. There's no reason that should ever occur. I would scrap the *entirety* of the builder changes and simply change line 200 of `_build/builder/CreateGitHubRelease.py` to `'prerelease': ("-" in self.buildSetup.appVersion)`.

As for the update checking code itself, the logic doesn't appear to be correct yet. If the latest version is a pre-release, it'll happily report that pre-release irrespective of the user's preference, and there's no exception for users already on a pre-release. The `...releases.latest.get()` call should be removed entirely, and you should instead loop through `...releases.get()`. Then:

    if item["prerelease"]:
        if eg.config.checkPreRelease or "-" in eg.Version.string:
            # Return this one and `break`.
    else:
        # Return this one and `break`.

Finally, I still prefer the "Always notify about new pre-releases" wording of the option. As is, if I were a user running a pre-release and unchecked "Include pre-releases in update check", I would expect not to receive any further update notifications for pre-releases, but that's not the case. The "Always notify" wording properly signifies how the option is actually applied.

blackwind:
One more thing: We should probably add an exception for `eg.Version.string.startswith("WIP-")`. There's no good way to determine if a release is newer or older than a WIP, so we shouldn't check for updates at all if that condition is true.

blackwind:
One more thing: We should probably add an exception for `eg.Version.string.startswith("WIP-")`. There's no good way to determine if a release is newer or older than a WIP, so we shouldn't check for updates at all if that condition is true. **EDIT:** Actually, I suppose you could compare `item["created_at"]` or `item["published_at"]` with `eg.Version.buildTime`.

blackwind:
Just a heads-up: I'll be introducing a ton of merge conflicts in the near-future. If you intend to finish this one, it might be wise to do so before that occurs.

topic2k:
Yes, i want to finish it, but currently have no time. So just go and make your changes, i will redo the PR later.

topic2k:
Ok, took some time and made some changes to the version check.

> , and there's no exception for users already on a pre-release. The ...releases.latest.get() call should be removed entirely, and you should instead loop through ...releases.get(). Then:
>
>
    if item["prerelease"]:
        if eg.config.checkPreRelease or "-" in eg.Version.string:
            # Return this one and `break`.
    else:
        # Return this one and `break`.

If i understand it correct, you want to search for prerelease version if the user is on a prerelease version, regardless if the option to check for prerelease versions is checked? Personally i wouldn't do that, but if you want, i'll add it.

blackwind:
Correct. We can safely assume that if the user is running a pre-release, they want to know when the next pre-release is available. This way, the user can freely jump between stable releases and pre-releases without modifying their options and always get update notifications for the channel they're on.

Based on the code at line 129 in `eg/Classes/CheckUpdate.py`, it looks like you might not be up-to-speed yet on WIPs. We have a new class of WIP now that's generated on every continuous integration build and can be found [here](https://ci.appveyor.com/project/blackwind/eventghost/build/artifacts). These WIPs have a version string like "WIP-2016.05.15-21.42.29", so your `if eg.Version.base == "WIP":` comparison will only apply to users running from source (which is what we want); meanwhile, CI WIPs aren't currently handled, and we'll need code for that as well.

There's no need to alert CI WIP users to new CI WIPs, but CI WIP users should be notified of new stable releases and pre-releases newer than their build. CI WIPs do contain a valid `eg.Version.buildTime`, so this is why I suggested comparing `item["created_at"]` or `item["published_at"]` with `eg.Version.buildTime` above. If it makes sense to do that same comparison across the board (and I can't think of any reason why not at the moment), you could eliminate the entirety of the version comparing code.

For now, this is the only specific note I have:

`eg/Classes/CheckUpdate.py`

* [line 129] Change to `if eg.Version.string == "WIP":`. I've just added code to [make `eg.Version.base` work as originally intended](https://github.com/EventGhost/EventGhost/commit/88826f415cf8f85ab9cfb3a0b2c29232c3523fb8), and the `eg.Version.base` for version string "WIP-2016.05.15-21.42.29" will be "WIP". We only want users running from source to receive this error message.

topic2k:
> We have a new class of WIP now that's generated on every continuous integration build and can be found here.

Ah, ok, now i understand (i think :D ).

blackwind:
Alright, we're almost there!

For clarity's sake, I still want this added:

> There should be an additional option (right below checkUpdateCtrl), disabled by default, called "Always notify about new pre-releases" so stable users can get pre-release notifications if they want them. **This option should be grayed out if the user disables the main check update option.**

If the user disables checkUpdate but still has checkPreRelease enabled, if it's not grayed out, he might assume that he'll still get pre-release notifications, which wouldn't be the case.

Beyond that, it's just a matter of cleaning up code that's no longer needed:

`_build/builder/Gui.py`

* Completely revert. There's still an import of `InvalidVersion` at the top that's no longer used, and there's still some horizontal spacer code that's no longer used. When you're done, this file should no longer appear in [files changed](https://github.com/EventGhost/EventGhost/pull/51/files) at all.

`eg/Classes/CheckUpdate.py`

* [lines 21, 189-250] Remove. If we need this code in the future, it'll be in the commit history.

blackwind:
Oh, one more thing:

`eg/Classes/Config.py`

* [line 90] Change to `checkUpdate = True`. Let's enable this by default.

topic2k:
crossing fingers... :)

blackwind:
I found one last issue, but I'm going to fix it myself. Merging.

topic2k:
ok, thank you.

PR #56, opened 2016-06-22
Add Wow6432Node key to Inno Setup detection

kdschlosser:
the original registry key didn't exist in my registry, I found another
that could be used. I added alongside the original one.

### DIFF
--- a/_build/builder/InnoSetup.py
+++ b/_build/builder/InnoSetup.py
@@ -95,7 +95,18 @@ def GetInnoCompilerPath():
         installPath = _winreg.QueryValueEx(key, "InstallLocation")[0]
         _winreg.CloseKey(key)
     except WindowsError:
-        return None
+        try:
+            key = _winreg.OpenKey(
+                _winreg.HKEY_LOCAL_MACHINE,
+                (
+                    "SOFTWARE\\Wow6432Node\\Microsoft\\Windows\\CurrentVersion\\"
+                    "Uninstall\\Inno Setup 5_is1"
+                )
+            )
+            installPath = _winreg.QueryValueEx(key, "InstallLocation")[0]
+            _winreg.CloseKey(key)
+        except WindowsError:
+            return None
     installPath = join(installPath, "ISCC.exe")
     if not exists(installPath):
         return None


kdschlosser:
i know you closed this. but i have compiled EG and i still had to add the other reg key into it because the one that is currently in the InnoSetup.py is not in my registry. i do not know as to why but it has nothing to do with what version of python i am trying to run eg with. as soon as i have that other key in it works fine.

it could be the result of possibly having installed a 64 bit version of python and then installed innosetup and then a 32 bit version of python afterwards, innosetup saw the 64 and only installed the 64 bit keys. i am not sure but that is my best guess.

topic2k:
I checked my registry and i too have the key `HKLM\SOFTWARE\WOW6432Node\Microsoft\Windows\CurrentVersion\Uninstall\Inno Setup 5_is1` and not `HKLM\SOFTWARE\Microsoft\Windows\CurrentVersion\Uninstall\Inno Setup 5_is1`, **but** python won't fail to get it:

```
Python 2.7.9 Stackless 3.1b3 060516 (default, Feb 21 2015, 11:54:09) [MSC v.1500 32 bit (Intel)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import _winreg
>>> key = _winreg.OpenKey(
...             _winreg.HKEY_LOCAL_MACHINE,
...             (
...                 "SOFTWARE\\Microsoft\\Windows\\CurrentVersion\\"
...                 "Uninstall\\Inno Setup 5_is1"
...             )
...         )
>>> installPath = _winreg.QueryValueEx(key, "InstallLocation")[0]
>>> print installPath
C:\Program Files (x86)\Inno Setup 5\
>>> exit()
```

So it looks like there is something wrong with your setup and/or registry.

blackwind:
I, too, have the Wow6432Node key. When a 32-bit process accesses `HKLM\SOFTWARE`, it's transparently redirected to `HKLM\SOFTWARE\Wow6432Node`, where it'll find the desired key. If this doesn't occur, you're not running the 32-bit version of Python, which may be because you've installed it alongside the 64-bit version and the 64-bit version comes first in your `%PATH%`.

My suggestion would be to pull the latest changes from the repo. Commit 2b5958a will prevent you from accidentally compiling with the 64-bit version.

blackwind:
That's the entire point of 2b5958a -- no 64-bit Python user will ever again get to the point where Inno Setup detection becomes an issue. Even if I were to reopen and merge this, the added code would never be executed.

kdschlosser:
i'm confused, because i still had this issue even thought i was compiling with a 32 bit version of python. i am not sure as to what chain of events is going on to not allow it to compile because it fails at the innosetup without that second reg key. it could be the path it could but something else i do not know. but if someone wanting to compile EG and has python 64 installed and then installs the 32 bit (if this in fact is what is causing the problem) could have the same issue as i. i am going to take one of my virtual machines and see if i can find out what the problem is.

blackwind:
If you can track down what's actually causing the problem and come up with a proper fix, I'd definitely be willing to merge that. This PR, though, is just a bandaid for the symptom that we'll need to remember to apply to any future `HKLM\SOFTWARE` access -- it may help you now, but this surely won't be the last time you experience the issue.

Ultimately, if you're certain you're running the 32-bit version, I fear @topic2k's hypothesis may be correct:

> So it looks like there is something wrong with your setup and/or registry.

kdschlosser:
It could be. but I am going to test this out on a virtual machine because i remember what I installed and in what order I installed it. just to see if it's a local computer problem or a specific order of installing things that caused it. It would figure I would be the one to have this kind of issue. LOL

topic2k:
@kdschlosser, can you check your `PATH` environment variable and look if the path of 32Bit-Python comes before ore after the 64Bit-Version? I don't know if this is causing the problem,it's just one thing i would check.

Issue #75, opened 2016-08-13
Modules I use in `site-packages` are missing


ThomasBott:
There seem to be a lot of site-packages missing in the new version.
I use psutil and paramiko for example.

topic2k:
None of the core functions or plugins use those packages. That's why they are missing. If you want to use them, you have to put them for example into the directory of the plugin using it or into `..\EventGhost\lib27\site-packages`.

topic2k:
It looks like there are issues with timezones and/or dst.

blackwind:
> None of the core functions or plugins use those packages. That's why they are missing.

Yep. Specifically, they were removed because of #20. You no longer have to petition Pako to include modules you want to use -- just [install Python 2.7](https://www.python.org/ftp/python/2.7.12/python-2.7.12.msi) (x86 version, as it has to match EventGhost), run `pip install paramiko psutil`, and EventGhost will pick them up.

ThomasBott:
hmm, I see, I must have missed that.
When I put something in site-packages, will it be removed with an update of eventghost or will it stay there?

Also: I can't get paramiko to work with this verison of EG:
```
import paramiko
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "C:\Python27\Lib\site-packages\paramiko\__init__.py", line 30, in <module>
    from paramiko.transport import SecurityOptions, Transport
  File "C:\Python27\Lib\site-packages\paramiko\transport.py", line 32, in <module>
    from cryptography.hazmat.backends import default_backend
  File "C:\Python27\Lib\site-packages\cryptography\hazmat\backends\__init__.py", line 7, in <module>
    import pkg_resources
  File "pkg_resources\__init__.pyc", line 68, in <module>
    t  File "pkg_resources\extern\__init__.pyc", line 60, in load_module
ImportError: The 'packaging' package is required; normally this is bundled with this package so if you get this warning, consult the packager of your distribution.
```
in python 2.7 natively it works just fine, any idea?

blackwind:
Incidentally, the commit for #81 fixes this as well, as we now bundle the full version of `pkg_resources`. Try beta2 when I release it later tonight. I'm able to successfully `import paramiko` now in my dev build.

> When I put something in site-packages, will it be removed with an update of eventghost or will it stay there?

It'll stay there, but I recommend you use `%PYTHONPATH%` instead. When we upgrade to a different version of Python, the `lib27` folder will no longer be loaded.

[...]

PR #79, opened 2016-08-13
Use the release name to check if it's newer than the installed version.

topic2k:
Currently the update check function checks if the release date is newer than the build time of the installed version. This is of course always true because the app has to be build before it is released. :smile:


blackwind:
Weird, didn't get a notification email for this, and didn't see it until just now, hours after I've already committed my own fix. Sorry about that. On the plus side, my fix fixes [this issue](https://github.com/EventGhost/EventGhost/issues/75#issuecomment-239647541) as well, so that's one less thing to worry about.

PR #92, opened 2016-08-16
Catch errors when adding a plugin

### LINKED ISSUE #90: Adding a plugin with invalid import call causes EventGhost to hang
If I add a plugin with a invalid import call (e.g. I import a plugin that's not available) doesn't throw an error but causes EG to hang.
Actually, it doesn't need to be an import command, anything that would throw an error before the plugin initializes causes EG to hang.

0.5.0 beta2

topic2k:
fixes #90
Please see this PR as a test as i don't know if it will break or influence anything else. At least for me it's working.

### DIFF
--- a/eg/Classes/Document.py
+++ b/eg/Classes/Document.py
@@ -189,7 +189,10 @@ class Document(object):
     def CmdAddPlugin(self):
         result = eg.AddPluginDialog.GetModalResult(self.frame)
         if result:
-            eg.UndoHandler.NewPlugin(self).Do(result[0])
+            try:
+                eg.UndoHandler.NewPlugin(self).Do(result[0])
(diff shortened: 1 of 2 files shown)


blackwind:
The `raise` concerns me, as this function is called from multiple places. I've settled on a better solution, but I'm going to merge this so you get the credit for tracking down the location of the problem.

blackwind:
Ended up going with an even better solution still -- keep the `raise` and handle it everywhere it needs to be handled.

topic2k:
> but I'm going to merge this so you get the credit for tracking down the location of the problem.

Thanks, but next time this is not needed, it's not my intention to do this for the credit. I'm glad if i can help, even if it just to inspire you to your own patch. :)

PR #94, opened 2016-08-17
Gracefully handle all possible plugin load errors

### LINKED ISSUE #84: Plugins are executed at startup rather than install
I have a plugin that is in development, and has some errors. In past versions of EG if the plugin was not installed it would not try to read the plugin file. This version now runs the plugin file at each startup. Not sure if this is by design, so I figured I would let you know.

blackwind:
Building on the work in #92, this will gracefully handle all possible plugin load errors. Whether you add an unparsable plugin, add a plugin with a broken `__init__()` or `__start__()`, load an unparsable plugin at startup, or load a plugin with a broken `__init__()` or `__start__()` at startup, you should never again end up with an empty EventGhost window and a log full of errors.

Resolves #84.


topic2k:
One file is missing: `eg/classes/LanguageEditor.py`
Line 78:
```
                    eg.pluginManager.OpenPlugin(plugin, plugin, ()).Close()
```
must be extended:
```
                    try:
                        eg.pluginManager.OpenPlugin(plugin, plugin, ()).Close()
                    except:
                        pass
```
Otherwise the language editor won't start.

blackwind:
Thanks, fixed.

PR #110, opened 2016-09-04
Finish the new "Add Event..." dialog

### LINKED ISSUE #40: Finish the new "Add Event..." dialog
It's already broken since at least version 0.4.1.1509 (the oldest i have/could run).

Clicking on a node to expand it, raises an exception:

```
Traceback (most recent call last) (0):
  File "D:\devel\EG-plum\eg\Classes\AddEventDialog.py", line 188, in OnExpanded
    self.tree.GetPyData(event.GetItem()).expanded = True
  File "D:\devel\EG-plum\eg\Classes\PluginModuleInfo.py", line 89, in __setattr__
    "%s has no attribute %s" % (self.__class__.__name__, name)
AttributeError: PluginInstanceInfo has no attribute expanded
```

Clicking "Ok" although raises an exception:

```
Traceback (most recent call last) (0):
  File "D:/devel/EG-plum/EventGhost.pyw", line 54, in <module>
    eg.Main()
  File "D:\devel\EG-plum\eg\__init__.py", line 83, in Main
    stackless.run()
  File "D:\devel\EG-plum\eg\Classes\MainFrame\__init__.py", line 1174, in OnCmdAddEventDialog
    label=label
  File "D:\devel\EG-plum\eg\Utils.py", line 161, in AssertWrapper
    return func(*args, **kwargs)
  File "D:\devel\EG-plum\eg\Classes\UndoHandler\NewEvent.py", line 53, in Do
    result = eg.actionThread.Func(ProcessInActionThread)()
  File "D:\devel\EG-plum\eg\Classes\ThreadWorker.py", line 64, in __call__
    self.returnValue = self.func(*self.args, **self.kwargs)
  File "D:\devel\EG-plum\eg\Classes\UndoHandler\NewEvent.py", line 34, in ProcessInActionThread
    for child in parent.childs:
AttributeError: 'NoneType' object has no attribute 'childs'
```

What should this dialog do? Simulate an event?
Is it ok to remove this function?

topic2k:
closes #40

### DIFF
--- a/eg/Classes/AddEventDialog.py
+++ b/eg/Classes/AddEventDialog.py
@@ -32,6 +32,10 @@ class Config(eg.PersistentData):
 class Text(eg.TranslatableStrings):
     title = "Add Event..."
     descriptionLabel = "Description"
+    noDescription = "<i>No description available</i>"
+    userEventLabel = "Manually enter event"
+    userEvent = "If an event is not in the list of available events," \
+                " it can be manually entered here."
 
 
 class AddEventDialog(eg.TaskletDialog):
@@ -39,12 +43,9 @@ class AddEventDialog(eg.TaskletDialog):
     def Configure(self, parent):
         global gLastSelected
         self.resultData = None
-        eg.TaskletDialog.__init__(
-            self,
-            parent,
-            -1,
-            Text.title,
-            style=wx.DEFAULT_DIALOG_STYLE | wx.RESIZE_BORDER | wx.THICK_FRAME
+        super(AddEventDialog, self).__init__(
+            parent=parent, id=wx.ID_ANY, title=Text.title,
+            style=wx.DEFAULT_DIALOG_STYLE | wx.RESIZE_BORDER
         )
         splitterWindow = wx.SplitterWindow(
             self,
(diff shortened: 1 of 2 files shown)


blackwind:
Just to give you a heads up, we're having [a major dispute about ownership](http://www.eventghost.net/forum/viewtopic.php?f=12&t=9068) over at the forum right now, which will probably lead to me leaving the project for good on the 15th. In the unlikely event that it ends up being resolved to my satisfaction, I'll be sure to get this reviewed and merged shortly thereafter.

ThomasBott:
@blackwind I don't want to raise a new discussion about this here, at the end you have to decide if you like to leave or to stay. Whatever your decision will be, I really appreciate your contribution to the project. It will be quite a loss if you leave and you should think about if one person you don't like, and that doesn't even change the code base, is reason enough to throw the last months of work in the bin. After all the Forum is not that important.

blackwind:
I appreciate the kind words. Working with the vast, vast majority of you has been and continues to be a pleasure.

Ultimately, with the goals we've set forth, we were never going to make it much further than v0.5 with zian standing in our way. Carson figured the work would speak for itself and that he'd come around after what we were about to accomplish; I knew better, having dealt with this type of personality many times, but chose to cross my fingers and hope for the best. This is what I get, I suppose, for ignoring my better judgment.

This time around, it's about so much more than the forum. As I note in the linked thread, if I can't get the access I need, I can't make the progress I'm here to make, and there's no sense in wasting even one more second of my time on a dead project.

Truly, it's a shame that one bad apple can spoil things for everyone, but it doesn't necessarily have to be this way. If you care as much as I do, please, make your voice heard -- there, not here. If it's the will of the community, it's entirely possible that he could do the right thing yet.

PR #111, opened 2016-09-08
Enabled flake8 check using travis

WoLpH:
Making all code fully pep8 compatible will be a lot more work but this is at least a start. I'll create an additional pull request for an autopep8 run. Not sure if we want to merge that (if someone is heavily modifying a branch right now it will conflict most likely)


WoLpH:
I've split the flake8 test so it tests every plugin separately which makes it easier to fix the plugins: https://travis-ci.org/WoLpH/EventGhost

That way we could make it ignore all non-fixed plugins and get an alert as soon as pep8 compliance is broken :)

Can you explain why this is labeled wontfix?

WoLpH:
Whoops... that explains why I was getting weird errors :)
Fixed now

topic2k:
Well, wontfix maybe isn't the right wording. I meant it more like wont include or wont merge. But i think i've misunderstood the whole thing: Travis will not change the files, instead it will show where things should be changed, right? Then i will change the label.

WoLpH:
Yes, exactly. It's simply a fully automated code quality test. Using these tests we can easily start converting plugins to adhere to the Python style guide and make sure they stay valid.

PR #112, opened 2016-09-08
Autopep8 to improve pep8 compliance

WoLpH:


topic2k:
Can you elaborate on using autopep8? If we will use it, i would like to have the pull request splittet into more smaller PR's: for the builder, eg, each plugin separate.

WoLpH:
The autopep8 is a one-off solution to improve the pep8 compliance. The pep8 standard is the default coding style for Python: https://www.python.org/dev/peps/pep-0008/

The changes are effectively whitespace only so it is slightly easier to check them using the ignore-whitespace feature of git(hub) diff: https://github.com/EventGhost/EventGhost/pull/112/files?w=1
It remains a lot though.

I can easily script it to create the separate pull requests for every plugin if you prefer but note that there will be 123 pull requests. (122 plugins + eg code)

topic2k:
> I can easily script it to create the separate pull requests for every plugin if you prefer but note that there will be 123 pull requests. (122 plugins + eg code)

Hmm, yeah, maybe just reduce it to eg, plugins and builder.

PR #116, opened 2016-11-16
[Webserver] Fixed incorrect "latin1" decoding

david-mark:
Query string parameters were incorrectly decoded as "latin1". Changed to "utf-8".

### DIFF
--- a/plugins/Webserver/__init__.py
+++ b/plugins/Webserver/__init__.py
@@ -1407,7 +1407,7 @@ class MyHTTPRequestHandler(SimpleHTTPRequestHandler):
         if remaining:
             queries = remaining.split("#", 1)[0].split("&")
             #print "queries =",queries
-            queries = [unquote_plus(part).decode("latin1") for part in queries]
+            queries = [unquote_plus(part).decode("utf-8") for part in queries]
             if len(queries) > 0:
                 event = queries.pop(0).strip()
                 if "withoutRelease" in queries:


topic2k:
I don't and haven't used the webserver plugin. Can others confirm the correctness of this fix? (Did not follow the forum about it.)

ThomasBott:
I'm testing this fix right now, I'll let you know next week if I noticed any problems.
In any case I will create a new Pull request, as there is more that changed in this plugin over the last two months than this one line and this should be merged as well.

david-mark:
https://en.wikipedia.org/wiki/Query_string

"All other characters are encoded as %HH hex representation with any non-ASCII characters first encoded as UTF-8 (or other specified encoding)"

UTF-8 is the default, unless another encoding is specified in the headers. So to be perfect will also need to check for other encodings.

david-mark:
...Personally, as discussed in the forum, I wouldn't bother to support anything but the default encoding of UTF-8. Simply specify in the documentation that it is the required encoding for HTTP requests. If you do bother to detect other encodings, just log an error.

david-mark:
I can certainly appreciate the value in testing, but this is one that is clearly wrong in the existing code. I had to add some silly Python workarounds just to pass non-ASCII characters on query strings. Tested extensively with Web browsers, Tasker, NodeJS, etc. This one string change is definitely the antidote as my workarounds look like this:

`
x = x.encode('latin1').decode('utf-8');
`

...And work like a charm. Just need to cut out the middleman here as most users are not going to come up with that (expect most won't even try). I barely did and came close to dumping EventGhost entirely due to all of the munged strings and exceptions that preceded settling on the above.

Cited the related standard in a previous post; I'm not even sure if it's possible to encode query string parameters in anything but UTF-8. Certainly it is possible for POST and PUT data by specifying the encoding, but the percentage of users that will do that AND specify 'latin1' is almost certainly nil. As mentioned, I would simply document that it is UTF-8 or bust (and at the moment it is undocumented that it is 'latin1' or bust).

ThomasBott:
@david-mark I don't see any problem with changing it to utf-8.
So I also don't mind if this PR get's merged so far. But I will create another PR with more changes very soon (about next weekend) that will also include this one most likely.

topic2k:
So, if it's ok for @david-mark i won't merge this PR and instead wait for the one from @ThomasBott.

@ThomasBott : will the PR include changes from the [forum](http://www.eventghost.net/forum/viewtopic.php?f=9&t=1663&start=210#p42980)?

ThomasBott:
Yes, all of them ;-)

david-mark:
Fine with me. Just glad to see it fixed. Thanks to all!

ThomasBott:
I tested it and can confirm, the change can be merged.

topic2k:
superseded by #134

PR #118, opened 2016-11-18
Bring back the -debug command line argument

topic2k:


ThomasBott:
What's the benefit of have the -debug switch over the new option feature "Log Debug Info"?

topic2k:
Sorry, didn't saw your comment earlier.

Using the switch will not store the value, so debug mode will only be turned on for this run of EG and you get debugging output before the UI is setup. Of course you can start EG, enable debug mode through the menu and restart EG. But what if you have an exception before the UI shows up?

ThomasBott:
ah, good point, thank you!

PR #119, opened 2016-11-20
Pass only decoded Unicode strings to TriggerEvent

david-mark:
Passing raw bytes received from UDP causes exceptions for non-Ascii characters. Change also allows for non-Ascii characters for the payload delimiter (previously caused an exception due to str() call).

See further explanation here:

http://www.eventghost.net/forum/viewtopic.php?f=4&t=8222&p=43750#p43750


david-mark:
Companion request so that sending encodes event names and payload. End result is that any Unicode string can make a UDP round trip without throwing exceptions or getting munged (and can use any unicode string for the delimiter).

https://github.com/EventGhost/EventGhost/pull/120

david-mark:
Note: may want to use eg.systemEncoding for instead of "utf-8". We don't get to specify this for HTTP, but UDP is open to whatever encoding we want to use.

PR #120, opened 2016-11-20
UTF-8 encode sent event name and payload Unicode

david-mark:
Brings send code into line with fixed receive.


david-mark:
Note: may want to use eg.systemEncoding for instead of "utf-8". We don't get to specify this for HTTP, but UDP is open to whatever encoding we want to use.

PR #124, opened 2016-11-22
Popup menu and hotkey for "Select All" in LogCtrl

topic2k:


ThomasBott:
This Change is the reason for #138

PR #125, opened 2016-11-24
FloatSliderCtrl widget

EventGhost-K:
Neat widget that allows for fine control over things like lights, thermostats.. could be used to tune a radio


topic2k:
I tried to use the FloatSliderCtrl, but it won't show up in the UI. Am i missing something?
Here's the code i used for testing:

    class TestAction(eg.ActionBase):
        def __call__(self, fval):
            print fval
    
        def Configure(self, fval=50.0):
            panel = eg.ConfigPanel()
            floatSlider = panel.FloatSliderCtrl(
                value=fval,
                minValue=10.4,
                maxValue=78.8,
                increment=0.2,
            )
            panel.sizer.Add(floatSlider)
            while panel.Affirmed():
                panel.SetResult(
                    floatSlider.GetValue()
                )

topic2k:
Parameter `increment` is missing in docstring.

topic2k:
floatsliderctrl

Shouldn't the current value be aligned under the slider?

EventGhost-K:
that's odd that the slider is showing up that way.

and also TY for noting the the increment not being in the docstring. i am going to add a parameter for it instead of just using kwargs.

and i think the removal of the "if vlabel or mlabel:" could be why the slider position may be off, i am going to go and fiddle with it now..


that position thing is really odd tho

EventGhost-K:
well i figured out what it was. I tested it with one size when i lined it up. never thinking to set it to a different size.... but i am pretty sure i did a panel resize. which is what is odd... but never the less... I am still working on alignment things. but I think i am going to have to set the pointer to always point at the decimal because of how it lines up with the min and max values. but I also took care of the issue with it having a gray background as well. because the widget is actually a panel with the wx Slider widget added to it I have the panel set the background colour to the system default window background. and the slider sets it's background colour by calling the panels method every time the EVT_PAINT is called the purpose to this is if you wanted to set the background colour you would set it for the panel. and the slider would grab it on a redraw. I am going to do the same thing for the foreground.

I do have one hang up tho that i realized I missed. if the EVT_SCROLL event. if someone by chance did GetEventObject() it would return the wxslider instance and if they called any methods on that instance it wouldn't be correct because i didn't override the slider instances methods..

so my question would be is this. is there any way to hook the EVT_SCROLL event that is generated from the slider and put the FloatSlider instance in its place??? I don't now if this would be easier to do then to override the methods in the slider instance

EventGhost-K:
OK. everything is looking good. just running the last couple of tests. still a little hung up on the EVT_SCROLL. so i am just going to override the slider methods. it's going to add a bucket load of code but it is what it is. i don't know how sizers handle getting and setting widget sizes. so i may have to ditch the easy way i did of adding the slider to a widget to have it handle the sizing. and also give me the space in the panel to make the labels. i may have to go with absolute sizing. as i already tried this and had all kinds of issues. but it was still added to a sizer at that point. so that is probably the issue.

and i believe i inadvertently added support for negative numbers as well. have to test it. ya gotta love it when accidents like that happen.

EventGhost-K:
OK i made the changes and also added proper event generation and handling. and other various things. I am not sure if the commit will just show up in this pull request or not. but please let me know.

and I would do the appveyor check but I am not sure how to use the thing. it keeps on telling me that windows 64 isn't supported. so possibly some instruction in that area would be helpful

Thanks

K

topic2k:
@EventGhost-K, did you pushed some commits by mistake? Do the package installer belong to FloatSliderCtrl? I don't think so, or?

topic2k:
For me the alignment of the value still wasn't working. What works for me, is replacing line 478 with
```
            if self.inverse:
                loc = sldRatio * (self.maxValue - value)
            else:
                loc = sldRatio * (value - minValue)
```

EventGhost-K:
I do apologize for that i didn't realize that if i did changed other files that it would affect this pull request...

I am new to gitub can't ya tell. LOL

I testing that thing every which way. what are you passing for args to the class?? i want to duplicate the problem

topic2k:
I'm testing it with:

        floatSlider1 = panel.FloatSliderCtrl(
            value=fval,
            minValue=10.4,
            maxValue=78.8,
            increment=0.2,
            style=wx.SL_LABELS
        )
        floatSlider2 = panel.FloatSliderCtrl(
            value=fval,
            minValue=10.4,
            maxValue=78.8,
            increment=0.2,
            style=wx.SL_LABELS | wx.SL_INVERSE
        )
        floatSlider3 = panel.FloatSliderCtrl(
            value=fval,
            minValue=10.4,
            maxValue=78.8,
            increment=0.2,
            style=wx.SL_LABELS | wx.SL_RIGHT
        )

EventGhost-K:
HAH! i was looking at my code and I had already accommodated for this. I didn't commit the most updated file. 

I think there is some kind of a glitch in wx or in the ConfigPanel I am not sure tho maybe you can check it as well.

But if you resize the Dialog it will size properly when dragging horizontal. but vertical does nothing. now i have an EVT_SIZE event already set up and in place. and I have been checking the incoming events and the size changes for a horizontal resize. but not the vertical. and i am unsure of how to handle this or if this is normal and does this to all widgets.

but I did notice that the ability to grow a column was only given wx.HORIZONTAL in the ConfigPanel

PR #129, opened 2016-12-14
New Widget: FloatSliderCtrl, mimics a wxSlider and allows floats

kdschlosser:
Custom wxSlider that handles floats and displays labels as well


kdschlosser:
Sorry about the removing of the pull request and adding it again. I am new to using Github and learned what branches are. So now you shouldn't get any commits that should not be there. I do apologize for this.

This version of the FloatSliderCtrl should be good to go (I hope). 
I added event creation to properly attach the FloatSliderCtrl instance to the event. This way if GetEventObject() was used it wouldn't return the actual wxSlider instance.

Also the "anomaly" in sizing is a wx thing and has nothing to do with EG. I do not understand why if I set a sizer to be vertical it will only report horizontal changes. and vice versa. I tried to use absolute positioning and sizing but because of the above mentioned it wouldn't work. I feel that unless I actually make a full blown widget and write the code to draw it, it will not function the way i want it to. It works as is but it always has the space for the labels present even if there are no labels displayed. 

I changed the string formatting for the float to be dynamic instead of absolute. it was set before to display no more then 3 decimal places. now it will look at the increment, the min value and the max value stripping off any trailing 0's then doing a split at the '.' and running max against the lengths of the 3. and who ever is largest is what gets used in the creation of the string format

i did find one bug. it's with the SL_AUTOTICKS if the ratio of range to increment has a large number of "steps"
Example: minValue=0.0, maxValue=10000.0, increment=0.0001
the wxSlider instance has a very large issue with drawing 10000000 tick marks.
i do not know where the breaking point in this is. but I would guess that it is machine dependent due to the users hardware. I do not know if there is a means to call the creation of the wxSlider and set a timeout period and if it times out to remove the SL_AUTOTICKS maybe you can advise on that. I thought about using threading for this. But I am unsure of how the system will react to hanging for something like 1/4 of a second.

topic2k:
> This version of the FloatSliderCtrl should be good to go (I hope).

:flushed: Sorry to say that, but i still have the problem with the wrong alignment of the current value and the slider:
[a short video](http://torsti.net/eg/eg-floatslider.mp4)
2016-12-18 at 17-55-17

And that's my test code:

    def Configure(self, fval=40.0):
        panel = eg.ConfigPanel()
        floatSlider1 = panel.FloatSliderCtrl(
            value=fval,
            minValue=10.4,
            maxValue=78.8,
            increment=0.2,
            style=wx.SL_LABELS
        )
        floatSlider2 = panel.FloatSliderCtrl(
            value=fval,
            minValue=10.4,
            maxValue=78.8,
            increment=0.2,
            style=wx.SL_LABELS | wx.SL_INVERSE
        )
        floatSlider3 = panel.FloatSliderCtrl(
            value=fval,
            minValue=10.4,
            maxValue=78.8,
            increment=0.2,
            style=wx.SL_LABELS | wx.SL_RIGHT
        )
        panel.sizer.Add(floatSlider1, 1, wx.EXPAND)
        panel.sizer.Add(floatSlider2, 1, wx.EXPAND)
        panel.sizer.Add(floatSlider3, 1, wx.EXPAND)

        while panel.Affirmed():
            panel.SetResult(floatSlider1.GetValue())

kdschlosser:
OK I am tired of the issues with wx and it's Sizers. because if i set a sizer to VERTICAL it will only report horizontal size changes. and if i set it to HORIZONTAL it reports only vertical size changes.

I am not not 100% sure what the issue is with it running in EG. but when i run it in it's own frame outside of the an eg config panel. it seems to work fine.

so i am just going to draw the whole thing instead of messing around with the sizers and spacers. this would have been far simpler if i was able to set the size of the widget to what i needed to fit the text also. because the slider only graphically changes size in one direction. the other is blank space and that is where my labels reside. but to no avail i am unable to draw on top of the widget. I am sure there is something I am missing. but it is a good thing i already did up the calculations to draw triangles as arrows from the center of the triangle. this is going to be useful in the creation on the thumb.

this will also give me the opportunity to fix the widget hanging with sl_autoticks and since the widget will be a panel with with the slider drawn on it I am hoping things will work properly

kdschlosser:
well this is where i am at. just as an update.

it's going very good. just have to add a polygon to the thumb and finish up the event generation

and the issue with the autoticks is fixed.
I also changed up how the page up and page down keys are going to work. i made it so that there are 2 different "tick sizes" large and small. and the page up and page down will do the large ticks and the up and down or just sliding will go per mark or increment.

here are some screen shots. you will see 2 controls. the top is the one i made. and the bottom is the original wxSlider

slider1
slider2

kdschlosser:
OK I shot out another commit.. Tho I know it is not done yet.. before I add anything that may be missing. or possibly not be functioning properly. I wanted bug test what I have so far before adding anything more to it. But it seems to be functioning pretty well. I have not had it throw a traceback and the alignment looks fantastic. This was an interesting bugger to tackle. But in the process I managed to make it faster then the stock wxSlider. and it also has no limitation on the number of tick marks. tho a huge number of them will cause a slow redraw but it will not lock up like the stock wxSlider does.

I also did a couple of things. like i reversed the scroll wheel when the slider is displayed vertically.. because the min value is at the top and the max at the bottom by default.. so if you scroll the wheel up it will go to the min where as if it's horizontal up is to max and down is to min...

I also implemented incorrect value being set. meaning that the value doesn't align with an increment.. this will automatically step to the nearest low step amount.

topic2k:
Getting better but still not working 100%  :neutral_face:

I use this [plugin ](https://gist.github.com/topic2k/f21362c634a13b4fdf5874fbc9f3ac19)for testing.
`minValue=10.4, maxValue=78.8, increment=0.2`

- When dragging the thumb with the mouse to the far right, i get:
    ```python
      File "D:\devel\EventGhost.topic2k\EventGhost\eg\Classes\FloatSliderCtrl.py", line 583, in GetThumb
        thumbPos = self.TickPosition[thumbPos][0]
    IndexError: list index out of range
    ```

- I can't focus the widget to use the keyboard to change the value/move the thumb
- with the mouse wheel i can change the value max to 52
- with the mouse wheel i can change the value to min 10.4 what is correct, but the thumb is not at the far left position
- if i use the mouse wheel to change the value and the value gets lower than 37.4, i can no more increase the value with the mouse wheel (until i drag the thumb above 37.4)

kdschlosser:
I know it has issues. this is a complicated bugger. and is my first fully custom widget... so a lot of patience is going to be needed... looks pretty good tho huh??? considering it is all rendered in my code. I think i have it identical in looks to the stock wxSlider.

I will have a look see at the issues you mentioned.. but this is going to go on the back burner until i get the other PR's finished

topic2k:
no need to hurry, i just need a nudge (prod? [you know what i mean?]) when i can try again. 😄

kdschlosser:
oh bogger... have to recommit...

kdschlosser:
OK The last commit should address the alignment issues. but also adds support for keyboard controls.

topic2k:
I used the same code for testing as in an [older comment](https://github.com/EventGhost/EventGhost/pull/129#issuecomment-267833085) .

- can't set the min/max values
- In inverse mode the value above the slider is not inverted.
- Using keys for moving slider doesn't work: After one key press the widget looses focus.
- pressing the number key 9 raises an exception 
```python
  File "E:\EventGhost-kdschlosser\eg\Classes\FloatSliderCtrl.py", line 658, in GetThumb
    thumbPos = self.TickPosition[thumbPos][1]
IndexError: list index out of range
```
unbenannt

kdschlosser:
send me the code that you are using to test this..

Also is the widget sizing properly now. as that was the thing i was concentrating on getting fixed.

topic2k:
The sizing seems ok, didn't notice a glitch.
For testing i modified the Test plugin:
```python
# -*- coding: utf-8 -*-

import eg

eg.RegisterPlugin(
    name="Test",
    author="Bitmonster",
    guid="{5533428D-57AB-4DAE-BA11-BB039DE0C4AA}",
    description=(
        "This plugin does nothing useful for end-users. It's purpose is to"
        "show developers some tricks in EventGhost"
    ),
)

import wx

class Test(eg.PluginBase):

    def __init__(self):
        self.AddAction(TestAction)

    def __start__(self, password=""):
        print "__start__", password

    def Configure(self, password=""):
        panel = eg.ConfigPanel()
        passwordCtrl = panel.PasswordCtrl(password)
        panel.sizer.Add(passwordCtrl)
        while panel.Affirmed():
            panel.SetResult(passwordCtrl.GetValue())

class TestAction(eg.ActionBase):

    def __call__(self, password):
        print repr(unicode(password))

    def Configure(self, fval1=40.0, fval2=40.0, fval3=40.0):
        panel = eg.ConfigPanel()
        floatSlider1 = panel.FloatSliderCtrl(
            value=fval1,
            minValue=10.4,
            maxValue=78.8,
            increment=0.2,
            style=wx.SL_LABELS
        )
        floatSlider2 = panel.FloatSliderCtrl(
            value=fval2,
            minValue=10.4,
            maxValue=78.8,
            increment=0.2,
            style=wx.SL_LABELS | wx.SL_INVERSE
        )
        floatSlider3 = panel.FloatSliderCtrl(
            value=fval3,
            minValue=10.4,
            maxValue=78.8,
            increment=0.2,
            style=wx.SL_LABELS | wx.SL_RIGHT
        )
        panel.sizer.Add(floatSlider1, 1, wx.EXPAND)
        panel.sizer.Add(floatSlider2, 1, wx.EXPAND)
        panel.sizer.Add(floatSlider3, 1, wx.EXPAND)

        while panel.Affirmed():
            panel.SetResult(
                floatSlider1.GetValue(),
                floatSlider2.GetValue(),
                floatSlider3.GetValue(),
            )
```

PR #130, opened 2016-12-14
Fixes incorrect reporting of the language in Windows

kdschlosser:
Changed using wx.local to ctypes to get the language.

Example of Bug:
Say you live in Germany but are English speaking. So you have the system locale set to Germany but the UI displayed language is in English. Using wx.locale would report back as de_DE even if the UI is set to en_US.

This fix reports the current displayed Windows GUI language. Not the system locale language.

Updated the code in the GetClosestLanguage() function so it only has one iteration and does any comparisons during the iteration.
Added a second line between function declarations in the Utils.py file to be pep8 compliant.
Removed an unreachable return statement.

### DIFF
--- a/eg/Utils.py
+++ b/eg/Utils.py
@@ -22,6 +22,7 @@ import sys
 import threading
 import time
 import wx
+from locale import windows_locale
 from CommonMark import commonmark
 from ctypes import c_ulonglong, windll
 from datetime import datetime as dt, timedelta as td
@@ -42,6 +43,7 @@ __all__ = [
 ]
 
 USER_CLASSES = (type, ClassType)
+
 
 class Bunch(object):
     """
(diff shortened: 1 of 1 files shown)


topic2k:
Hmm, i can't replicate the problem:

---
de_de-1
de_de-2

---
en_us-1
en_us-2

---

Please do separate pull requests for functional code changes and pep8 code changes.

kdschlosser:
Ya know I have been testing this and I cannot seem to get it to reproduce either.

By chance was the last release of EG running a non unicode version of wx??
If it was then it could have been the non unicode wx. And since wx has been upgraded and wx now only has one type and it's unicode. that could have solved the problem or it could be that settings have to be a specific way. 

I really don't remember. I do remember running across the issue. If you read #123, I bring up this change, someone makes a comment about it in a manner that would make you think they had experienced the same issue as well.

But at any rate I do recall you stating the possibility of migrating away from wx in favor of some kind of web browser control instead. So this will satisfy the means to get the proper displayed language without the use of wx. So keep it on file somewhere. Could prove to be useful.

PR #132, opened 2016-12-14
Change the MainFrame to use wx.agw.aui instead of wx.aui

kdschlosser:
Changed the AUI from wx.aui to wx.lib.agw.aui
Changed the ToolBar to an AuiPane
Updated the code for the adding of the treeCtrl pane as well as the logCtrlPane
Removed the "&" from the tool tips (short tool help) for the ToolBar items
Removed the double entries for setting the captions on the treeCtrl pane and the logCtrl pane

This update allows for the displaying of docking guides as well as a more fine grained control over the AUI for future enhancements.

### DIFF
--- a/eg/Classes/MainFrame/__init__.py
+++ b/eg/Classes/MainFrame/__init__.py
@@ -20,7 +20,7 @@ import inspect
 import re
 import types
 import wx
-import wx.aui
+from wx.lib.agw import aui
 from collections import defaultdict
 from os.path import join
 
@@ -59,6 +59,7 @@ ID = defaultdict(wx.NewId, {
 })
 
 Text = eg.text.MainFrame
+
 
 class Config(eg.PersistentData):
     position = (50, 50)
@@ -104,12 +105,14 @@ class MainFrame(wx.Frame):
         )
         self.SetMinSize((400, 200))
         document.frame = self
-        auiManager = wx.aui.AuiManager(self, wx.aui.AUI_MGR_DEFAULT)
-        self.auiManager = auiManager
+        self.auiManager = auiManager = aui.AuiManager()
+        auiManager.SetManagedWindow(self)
 
         self.logCtrl = self.CreateLogCtrl()
-        self.corConst = self.logCtrl.GetWindowBorderSize()[0] + \
+        self.corConst = (
+            self.logCtrl.GetWindowBorderSize()[0] +
             wx.SystemSettings.GetMetric(wx.SYS_VSCROLL_X)
+        )
         self.treeCtrl = self.CreateTreeCtrl()
         self.toolBar = self.CreateToolBar()
         self.menuBar = self.CreateMenuBar()
@@ -134,9 +137,9 @@ class MainFrame(wx.Frame):
         self.ratio = Config.ratio
         self.Bind(wx.EVT_SIZE, self.OnSize)
         self.Bind(wx.EVT_MOVE, self.OnMove)
-        self.Bind(wx.aui.EVT_AUI_PANE_CLOSE, self.OnPaneClose)
-        self.Bind(wx.aui.EVT_AUI_PANE_MAXIMIZE, self.OnPaneMaximize)
-        self.Bind(wx.aui.EVT_AUI_PANE_RESTORE, self.OnPaneRestore)
+        self.Bind(aui.EVT_AUI_PANE_CLOSE, self.OnPaneClose)
+        self.Bind(aui.EVT_AUI_PANE_MAXIMIZE, self.OnPaneMaximize)
+        self.Bind(aui.EVT_AUI_PANE_RESTORE, self.OnPaneRestore)
         self.UpdateViewOptions()
         self.SetSize(Config.size)
         eg.Bind("DocumentFileChange", self.OnDocumentFileChange)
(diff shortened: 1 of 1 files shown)


eg/Classes/MainFrame/__init__.py
```diff
@@ -343,6 +337,10 @@ def FuncWrapper(dummyEvent):
         menu.AppendSeparator()
         Append("Disabled", "\tCtrl+D", kind=wx.ITEM_CHECK)
 
+        # settings menu
+        settingsMenu = eg.SettingsMenu(ID)
```
topic2k: `eg.SettingsMenu` doesn't exists.

kdschlosser:
Sorry about that. I was working on something else and forgot that I had that in there. It has been removed.

topic2k:
I like the idea of having dockable panes and toolbars, but i'm not sure if we should implement the use of agw.aui. I would like to hear what others think, as i found some behaviour i don't like. I made a short video of it: http://torsti.net/eg/eg-aui.mp4
Maybe it's only me having a problem with it. If other's want the change too, then i'm happy to merge the PR.

kdschlosser:
the current state of eg allows docking.  it just doesn't show the docking guides. and the toolbar was set as a frame toolbar and not a pane.  and it doesn't have the full feature set that the agw aui has.

kdschlosser:
and that issue you mentioned... you have to float the window completely. meaning let go of the mouse button LOL

then double click on the caption bar of the floating pane. I am thinking this is because the location you are trying to "float" it back into is it's default spot. and if that is where you want it to go back into then you have to double click the caption instead..

I am sure there is a means to work around this. but using the agw.aui offers a whole lot more "flare"
I have created a settings pane that uses agw.foldpanelbar It's still in development but it works and it allows you to control just about every aspect of the agw.aui. but it also has an added bunos. it's not limited to just that. it could be used as a foundation for plugin settings (instead of dialogs), or the core settings. I am not a huge fan of the whole dialog thing. i hate windows popping up all over the place. having the option to either float or dock them is a great feature. but with the current use of the dialogs you can't make them into panes. and I don't know about you but also having to have the plugin completely restart for a settings change seems kind of excessive.

topic2k:
Hej K, it's a while since i looked at it. I think i found the solution to "my" 😉 problem. Replace
```python
        self.auiManager = auiManager = aui.AuiManager()
        auiManager.SetManagedWindow(self)
```
with
```python
        self.auiManager = auiManager = aui.AuiManager_DCP(self)
```
I only did a quick check, but seems to be working for me.

kdschlosser:
I have never checked out AuiManager_DCP. I don't know what may be different then the straight AuiManager.. I will read up on it and see what the difference is..

topic2k:
From the source code: `A class similar to :class:'AuiManager' but with a Dummy Center Pane (**DCP**).`

kdschlosser:
thats right... 

I thihnk I know what is causing the issue.. i followed the original cod and keys the docking and floating and where the panes got set to the same as the original..

but the way the aui works is one pane always has to stay with the parent frame.. and it has to have the Center() designation as well as not be able to be floated

the current setup puts the log as the left and it can't be floated. and the tree to the center and it can be..

so we can test out swapping the positions around. make the log l center and the tree right.. and make the tree floatable and the log not.


I bet that will fix your issue without the use of the dummy center pane... because we really do not want an empty frame for EG and any given time.. locking the log to it would be ideal because of the size that the log usually is set to.. so the menubar can be fully displayed.


so what we need to do is change to this

        pane = aui.AuiPaneInfo()
        pane.Name("tree").Caption(" " + Text.Tree.caption).Right()
        pane.MaximizeButton(True).CloseButton(False).MinSize((100, 100))
        pane.Floatable(True).Dockable(True)
        self.auiManager.AddPane(treeCtrl, pane)
        self.auiManager.Update()


and this

        pane = aui.AuiPaneInfo()
        pane.Name("logger").Caption(" " + Text.Logger.caption).Center()
        pane.MaximizeButton(True).CloseButton(False).MinSize((100, 100))
        self.auiManager.AddPane(logCtrl, pane)
        self.auiManager.Update()

topic2k:
Yes, that's an option. I already thought about that but didn't say it because i thought you wanted all be dockable/floatable. Then i would suggest to fix the logctrl
```python
        pane = aui.AuiPaneInfo()
        pane.Name("logger").Caption(" " + Text.Logger.caption).Center()
        pane.MaximizeButton(False).CloseButton(False).MinSize((100, 100))
        pane.Floatable(False)

```

That would satisfy me and make me happy 🤣

kdschlosser:
all fixed.... it was as i thought..

but there were a couple of other things i tackled in the last commit as well.. the proportional resizing needed to be converted to work with the fact that the log is always docked and is center pane.

but also I added some checking for EG version based on the Config.perspective so if coming from an older version of EG it will reset the toolbar this is due to the fact that the toolBar information is not in the saved perspective.. so it doesn't show it properly.

but it also checks the docked state of the tree then docks the tree and the log sets the positions and floating flags then refloats the tree if it was floated.

doing my best here to make this stuff appear as seamless to the user as possible. until they go to use it..

The big jump is when we add some kind of an adjustment pane for setting colors and all the various aui styles and the user can select what they want..

I have something all set for that as well. it just needs to be polished. giving the user the ability to customize is a good thing.. another cool thing is it allows for exporting and importing of "themes" so they can be traded between people.

topic2k:
Seamless sounds good 😃  
Sorry to say that i found another small thing. Maybe it comes from my testing, but if i have the following perspective info in my config.py, i still can drag the logctrl and make it float.

`            perspective = u'layout2|name=logger;caption=Log;state=4196348;dir=4;layer=0;row=0;pos=0;prop=100000;bestw=300;besth=300;minw=100;minh=100;maxw=-1;maxh=-1;floatx=-1;floaty=-1;floatw=-1;floath=-1;notebookid=-1;transparent=255|name=tree;caption=Configuration;state=4196348;dir=5;layer=0;row=0;pos=0;prop=100000;bestw=100;besth=100;minw=100;minh=100;maxw=-1;maxh=-1;floatx=-1;floaty=-1;floatw=-1;floath=-1;notebookid=-1;transparent=255|name=toolBar;caption= Toolbar;state=67379964;dir=1;layer=10;row=0;pos=0;prop=100000;bestw=444;besth=26;minw=-1;minh=-1;maxw=-1;maxh=-1;floatx=-1;floaty=-1;floatw=-1;floath=-1;notebookid=-1;transparent=255|dock_size(4,0,0)=601|dock_size(5,0,0)=102|dock_size(1,10,0)=28|'
`

And set `MaximizeButton()` for the logctrl to `False`?

ouh, problem:   :fearful:
- dragging the config pane onto the log pane, so that you have a notebook
- now minimize EG
- show EG
- the notebook is gone and if you drag the sash between the panes you get exceptions.

kdschlosser:
as far as that specific perspective you are using. it has to be from an earlier version of this specific update.. because toolbar is in the perspective... before this update the toolbar was not a pane therefor was not in the perspective.. you have to remember people coming from any other version of EG will not have the toolbar in the perspective.. and i look for toolbar in it. and if it's not then i will go ahead and modify things.. and how aui works is you create all of the panes and then load the perspective with will change all of the values you have set for the panes. is they differ for some reason. and in earlier version of this PR we didn't lock the log to the main frame. and you are using a perspective from then.

I am going to check on the other things as well in a little bit.

kdschlosser:
I would like to do some house keeping where are we at with this?

topic2k:
uh, can't remember. I checked out the branch and still have the problem as in my previous comment. But that is a wx error.

kdschlosser:
I don't remember either. I really do like the idea of having this added. and I would love to get it working properly. I thought the last commit i did solved the issues with it. I know I am being a pain. but if you could have a gander at it again and see if you can replicate an issue with it and tell me what it is we can go from there. I am pretty sure that i fixed the docking problem.

PR #137, opened 2016-12-22
PluginInstall: added backup, GUID check, better user notification, pretty print the info.py

kdschlosser:
Adds more informative message when running the plugin installer. Messages for Upgrading plugin, Downgrading plugin, Overwriting plugin, Plugin exists but the name is different as well as Install plugin

Adds a GUID check when exporting, if one does not exist a dialog will ask to create one. if one is created it will modify the __init__.py and add a newly generated one. It WILL NOT allow a plugin to be exported without a GUID.

Adds a backup routine for the old version of the plugin if one is installed. it will append the old version to the file name when backing up

Adds a file name suggestion when exporting that has the version of the plugin included in the file name

Adds exception handling for removing the old plugin directory if the plugin directory resides in the eg.corepluginDir it will post a message in the log letting the user know they have to go and manually remove it. otherwise raise the exception


kdschlosser:
not sure about the code climate thing. Unless I break apart the methods/functions to reduce the number of "if" statements and operators used in a single method It will always fail. But to be honest with you I think that making 5 methods/functions to do the same work as one because the code is not shared makes it a lot more confusing to read and follow. The only reason to make a method/function is because the code needs to be shared. Either with the code in the class/module or with an external request. Other then that there really is no reason for it, if there are 1000 "if" statements and another 1000 operators, It doesn't matter, because the code housed in the method/function does one thing and only one thing. And therefor really has no purpose being separated.
And that's my rant for the day.

:-D

topic2k:
Don't care to much about codeclimate status for now. It needs to be fine grained.

kdschlosser:
cool. So i don't have to go all bonkers trying to get it to pass those tests. I do hate the ugly red X tho.

lol

topic2k:
Adding a guid might result in a corrupt plugin. Consider this:

    ICON = (
        "iVBORw0KGgoAAAANSUhEUgAAA"
        "S0dEAP8A/wD/oL2nkwAAAAlwSFlz"
    )
    
    eg.RegisterPlugin(
        ...
        icon = ICON,
        ...
    )

When the plugin info has no guid and on export i choose to automatically add a guid, it will result in wrong plugin info:


    ICON = (
        "iVBORw0KGgoAAAANSUhEUgAAA"
        "S0dEAP8A/wD/oL2nkwAAAAlwSFlz"
    )
    
    eg.RegisterPlugin(
        ...
        guid=u'{ ... }',
        ...
        icon = 'ICON'
    )

kdschlosser:
I did consider that... and accounted for it as well :-D.

    if icon.endswidth(','):
        icon += unicode(iconLine.strip()[1:-1])
        icon = icon[:-1]
    if icon[:1] in ("'", '"'):
        icon = repr(unicode(icon[1:-1]))

topic2k:
It doesn't work, believe me, i've tested it. Maybe you made some code changes and didn't commit them?

With the following patch it works:

diff --git a/eg/Classes/PluginInstall.py b/eg/Classes/PluginInstall.py
index 9128049..413a0d8 100644
--- a/eg/Classes/PluginInstall.py
+++ b/eg/Classes/PluginInstall.py
@@ -21,6 +21,7 @@ import ast
 import base64
 import compileall
 import os
+import re
 import shutil
 import tempfile
 import wx
@@ -89,7 +90,7 @@ def InfoField(field, attr, tab=''):
 
 def FindStop(nextLines, icon):
     for j, nextLine in enumerate(nextLines):
-        ic = nextLine.startswith('    icon=') or nextLine.startswith('\ticon=')
+        ic = re.match(r"^(\s*)icon(\s*?)=", nextLine)
         if icon and ic:
             icon = nextLine[nextLine.find('=') + 1:].strip()
             if icon.startswith('('):
@@ -100,15 +101,15 @@ def FindStop(nextLines, icon):
                         break
                     icon += unicode(iconLine.strip()[1:-1])
             else:
-                if icon.endswidth(','):
+                if icon.endswith(','):
                     icon = icon[:-1]
                 if icon[:1] in ("'", '"'):
                     icon = repr(unicode(icon[1:-1]))
 
         if nextLine.rstrip() == ')':
-            return j, repr(icon)
+            return j, icon
 
-    return None, repr(icon)
+    return None, icon
 
 
 class PluginInstall(object):

kdschlosser:
I didn't know you tested it.. and that is did in fact not work properly... I may have missed testing that one bit of it. But I did think about it and try to account for it. so that's a +. LOL

I know this whole thing could be improved upon with the use of Regex. I am not proficient enough in it
.So how about this.
The Idea is sound Yes?
Run with it. Clean it up by using Regex. I know the whole process of getting the eg.RegisterPlugin from the __init__ could be simplified a whole lot with the use of it.. 

Make it better. 

I would like to know what you think about the rest of the changes as well?.

topic2k:
I like the idea of removing old data, but i would never do it without asking the user.

kdschlosser:
I really have no clue as far as the Regex thing. it's a bloody mystery to me. and the fact that there is god knows how many different "flavors" of Regex on top of it. and i haven't been able to find any documentation on it that is not geared at someone already knowing how to use regex,

topic2k:
Yeah, regex is sometimes dificult to understand. I found [this site](https://regex101.com/#python) helpfull for testing and understanding regex.

kdschlosser:
Oh hey I forgot to toss a line in here.. because it doesn't show you if i did a commit.. but i did one...

topic2k:
What i did: exported the plugin "Sceneo TVcentral". When i then import that egplugin-file i get an SyntaxError. Looking at the created info.py i see the problematic. The description contains an `'` which i think irretates your `InfoFild `function.

kdschlosser:
no worries. I will have a look see at it.

kdschlosser:
all fixed.. it was caused by not converting \n to \\n. 
 and I also looked at the code and was thinking.... Boy am I an idiot.. so i shortened it up some. and killed some repeat code. so it's a little easier to follow.

It really stinks about the GUID thing.. but there are just way to many variables to account fore without making a large chunk of code just for that...

i have seen where the version parameter was passed via an attribute
icon as well...

i have seen the description set as a docstring in the __init__ file and then the __doc__ passed to the description parameter

complete loony tunes if ya ask me.

so as a replacement i just added a simple dialog. to spit out a guid. and instruction on where to enter it.. but to make it really easy so the user could copy and paste the thing. since classic wx doesn't allow for the passing of Text styles to a dialog.. instead of amking a whole dialog.. i just went and poked it with a stick. to give me the wxTextCtrl instance it was using.. and set the thing to read only directly.

eg/Classes/PluginInstall.py
```diff
@@ -24,7 +24,9 @@
 import shutil
 import tempfile
 import wx
+import re
```
topic2k: importet module `re` isn't used,

eg/Classes/PluginInstall.py
```diff
@@ -96,11 +144,38 @@ def Export(self, pluginInfo):
         #dialog.Destroy()
         #if result == wx.ID_CANCEL:
         #    return
-        filename = os.path.basename(pluginInfo.path)
+
+        if pluginInfo.guid == pluginInfo.englishName:
```
topic2k: the check must be against `pluginInfo.pluginName`. For example:
```
pluginInfo.guid: AIRT
pluginInfo.englishName: Advanced IR-Transceiver
pluginInfo.pluginName: AIRT
```

eg/Classes/PluginInstall.py
```diff
@@ -131,6 +206,28 @@ def GetPluginData(self, pluginInfo):
...
+        backupPath = os.path.join(
+            eg.folderPath.ProgramData,
+            'EventGhost',
+            'PluginBackups'
+        )
+        if not os.path.exists(backupPath):
+            os.makedirs(backupPath)
+
+        eg.PrintNotice('Creating backup of plugin.')
+        eg.PrintNotice('Writing ' + os.path.join(backupPath, backupName))
```
topic2k: I would prefer to see an info icon in the log instead of the yellow triangle. It looks to me like a warning that something could be wrong. (eg.Print instead of eg.PrintNotice). Just my opinion.

eg/Classes/PluginInstall.py
```diff
@@ -146,38 +243,78 @@ def Import(self, filepath):
...
+
+                self.BackupPlugin(info)
+                try:
                     shutil.rmtree(info.path, False)
-            dstDir = os.path.join(eg.localPluginDir, os.path.basename(basePath))
+                except WindowsError:
+                    if info.path.startswith(eg.corePluginDir):
+                        eg.PrintNotice(Text.pluginRemoveError + info.path)
+                    else:
+                        raise
```
topic2k: Maybe instead of (re-)raising an exception show a dialog that the removal of the plugin failed and the (new/updated) plugin won't be installed. (could it even be that the plugin might now be unusable?)

kdschlosser: the reason for the reraise is because of some unknown problem. permissions error or what have you.

We do know that with windows 10 it will not allow for a program to modify anything in the program files folder. so if EG goes to try and remove the old plugin and it is located in the core plugins directory which is normally in the program files folder. there is a really good change that is why an error has occurred.. tho i do not specifically state this due to installation location and things of that nature. but by checking to see if the plugin was housed in the core plugin dir. the chances are pretty good that it has an issue removing it.. so letting the user know that they have to do it manually I thought would be the best thing..

now if the plugin is not in that folder.. then I have not a clue what the issue could have been. and reraising is what i thought would be best. originally there was no message dialog explaining any kind of an error. it would just throw an exception.

so this is one step better... 

and as far as the icon thing... i do agree with you on this.. and I think there should be a eg.PrintInfo for this purpose. and print should have no icon what's so ever. but as it is set up now print has the nice information icon. now. we can change that up if you like it's pretty simple to do. I am personally for it..
but.. i also think it should be one step further.. and there should be some kind of separation between plugin PrintError, PrintNotice and PrintInfo..    VS a core EG one. and an icon is a good way of giving that. I have thought about this in the past... but I do think that at the very least migrating the information icon away from stdout and put it into a new method in eg.Log is a good choice. this way it will also get a time stamp where as the print does not... and also i think that doing a print is really for debugging purposes and nothing more, and has no need for an icon.


and if memory serves the english name is what the guid gets set to.. I will have to check.. and the pluginName is the class name.. I think... again i will have to look

heh sorry forgot to remove the import re.

PR #140, opened 2016-12-27
Ask to delete config data of plugin if uninstalled

kdschlosser:
When removing a plugin eg.config is scanned for any eg.PersistantData the plugin may have set. and if it finds any prompts the user if they would like to delete this data. If yes the data is marked for deletion. If the user does not save their EventGhost session all of the marks for deletion are removed. It also handles the ImportAll() when the debug level is set from a command line.

### DIFF
--- a/eg/Classes/PluginItem.py
+++ b/eg/Classes/PluginItem.py
@@ -60,21 +60,6 @@ class PluginItem(ActionItem):
     def AskCut(self):
         return self.AskDelete()
 
-    def GetPersistantData(self):
-        info = self.info
-        if info.path.startswith(eg.corePluginDir):
-            configName = "CorePluginModule"
-        else:
-            configName = "UserPluginModule"
-
-        if hasattr(eg.config.eg, configName):
-            config = getattr(eg.config.eg, configName)
-        else:
-            config = None
-
-        if config and hasattr(config, info.pluginName):
-            return 'eg.%s.%s' % (configName, info.pluginName)
-
     def AskDelete(self):
         actionItemCls = self.document.ActionItem
 
@@ -94,16 +79,15 @@ class PluginItem(ActionItem):
         if not TreeItem.AskDelete(self):
             return False
 
-        config = self.GetPersistantData()
-
-        if config:
+        if self.info.persistentData.config:
             answer = eg.MessageBox(
                 eg.text.General.deleteConfigMessage,
                 eg.text.General.deleteConfigCaption,
                 wx.YES_NO | wx.ICON_QUESTION
             )
             if answer == wx.ID_YES:
-                eg.config.MarkForDelete(config)
+                self.info.persistentData.Delete(True)
+
         return True
 
     @eg.AssertInActionThread
@@ -114,7 +98,6 @@ class PluginItem(ActionItem):
             info.Close()
             info.instance.OnDelete()
             info.RemovePluginInstance()
-
         eg.actionThread.Call(DoIt)
 
         ActionItem.Delete(self)
@@ -175,10 +158,6 @@ class PluginItem(ActionItem):
         """
         # if the Configure method of the executable is overriden, we assume
         # the item wants to be configured after creation
-        config = self.GetPersistantData()
-        if config:
-            eg.config.AddUndoData(config)
-
         return (
             self.executable.Configure.im_func !=
             eg.PluginBase.Configure.im_func
--- a/eg/Classes/PluginItem.py
+++ b/eg/Classes/PluginItem.py
@@ -81,7 +81,7 @@ class PluginItem(ActionItem):
 
         if self.info.persistentData.config:
             answer = eg.MessageBox(
-                eg.text.General.deleteConfigMessage,
+                eg.text.General.deleteConfigMessage % self.info.name,
                 eg.text.General.deleteConfigCaption,
                 wx.YES_NO | wx.ICON_QUESTION
             )
--- a/eg/Classes/PluginItem.py
+++ b/eg/Classes/PluginItem.py
@@ -60,6 +60,21 @@ class PluginItem(ActionItem):
     def AskCut(self):
         return self.AskDelete()
 
+    def GetPersistantData(self):
+        info = self.info
+        if info.path.startswith(eg.corePluginDir):
+            configName = "CorePluginModule"
+        else:
+            configName = "UserPluginModule"
+
+        if hasattr(eg.config.eg, configName):
+            config = getattr(eg.config.eg, configName)
+        else:
+            config = None
+
+        if config and hasattr(config, info.pluginName):
+            return 'eg.%s.%s' % (configName, info.pluginName)
+
     def AskDelete(self):
         actionItemCls = self.document.ActionItem
 
@@ -78,6 +93,17 @@ class PluginItem(ActionItem):
             return False
         if not TreeItem.AskDelete(self):
             return False
+
+        config = self.GetPersistantData()
+
+        if config:
+            answer = eg.MessageBox(
+                eg.text.General.deleteConfigMessage,
+                eg.text.General.deleteConfigCaption,
+                wx.YES_NO | wx.ICON_QUESTION
+            )
+            if answer == wx.ID_YES:
+                eg.config.MarkForDelete(config)
         return True
 
     @eg.AssertInActionThread
@@ -88,6 +114,7 @@ class PluginItem(ActionItem):
             info.Close()
             info.instance.OnDelete()
             info.RemovePluginInstance()
+
         eg.actionThread.Call(DoIt)
 
         ActionItem.Delete(self)
@@ -148,6 +175,10 @@ class PluginItem(ActionItem):
         """
         # if the Configure method of the executable is overriden, we assume
         # the item wants to be configured after creation
+        config = self.GetPersistantData()
+        if config:
+            eg.config.AddUndoData(config)
+
         return (
             self.executable.Configure.im_func !=
             eg.PluginBase.Configure.im_func
--- a/eg/Classes/Text.py
+++ b/eg/Classes/Text.py
@@ -24,6 +24,8 @@ from eg.Utils import SetDefault
 
 class Default:
     class General:
+        deleteConfigMessage = "Do you want to remove all configuration data?"
+        deleteConfigCaption = "Delete Configuration Data"
         configTree = "Configuration Tree"
         deleteQuestion = "Are you sure you want to delete this item?"
         deleteManyQuestion = (
--- a/eg/Classes/Text.py
+++ b/eg/Classes/Text.py
@@ -24,7 +24,10 @@ from eg.Utils import SetDefault
 
 class Default:
     class General:
-        deleteConfigMessage = "Do you want to remove all configuration data?"
+        deleteConfigMessage = (
+            "Do you want to remove all configuration data\n"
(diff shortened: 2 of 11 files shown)


kdschlosser:
this was a little bit of a pain to do. I hope i managed to cover all of the bases and any possible way config data would get set and removed. Oh I also covered my bases in the department of not leaving a parent class vacant. if it's empty it removes the parent class as well.

It appears I also did a little better in the codeclimate department this time. LOL

I tried to test it every way I could possibly think of. if ya find something lemme know

topic2k:
Can you tell me how to test it? Never used persistant data.

kdschlosser:
install a plugin that has persistent data. and load your config.py file. save and close EG..

remove the plugin then save and close EG.

you will notice the config data doesn't go away.

use my modifications.. then do the same thing..

now when you use the -debug ImportAll() happens.. so watch that also.. because it will add all of the persistent data for all of the plugins that use it..

and watch because it will remove them if they didn't already exist in the config..
unless..you add a plugin while in debug then it will leave it there. it also create an Undo" set just in case you remove the pluguin then don't save it will put it back into place,,,

but it also has to look for an empty parent class as well.. because if there are say no userplugins installed. you can't have an empty class.. so it will delete that as well if it is empty

topic2k:
Seems like something goes wrong when running EG with -debug.
My `config.py` looks like:
```
autoloadFilePath = u'D:\\topic\\Documents\\Neuer Ordner\\speech.egtree'
checkPreRelease = False
checkUpdate = True
confirmDelete = True
hideOnClose = False
hideOnStartup = False
language = 'en_EN'
lastUpdateCheckDate = (2016, 12, 31)
limitMemory = False
limitMemorySize = 8
propResize = True
refreshEnv = False
showTrayIcon = True
useFixedFont = False
version = 'WIP'
class eg:
    class Classes:
        class AddActionDialog:
            position = None
            size = (550, 400)
            splitPosition = 210
        class AddEventDialog:
            position = None
            size = (550, 400)
            splitPosition = 210
        class AddPluginDialog:
            collapsed = set([])
            lastSelection = None
            position = None
            size = (640, 450)
            splitPosition = 240
        class Document:
            expanded = set([0, 1, 2, 3])
            firstVisibleItem = []
            guid = '{FA510517-F5F9-4519-A6CC-E777A0175D52}'
            selection = []
            time = '1483205484.24'
        class HtmlDialog:
            position = None
            size = (400, 300)
        class LanguageEditor:
            language = None
            position = (50, 50)
            size = (700, 433)
            splitPosition = 244
        class MainFrame:
            expandOnEvents = False
            indentLog = True
            logTime = False
            perspective = u'layout2|name=logger;caption=Log;state=4196348;dir=4;layer=0;row=0;pos=0;prop=100000;bestw=300;besth=300;minw=100;minh=100;maxw=-1;maxh=-1;floatx=-1;floaty=-1;floatw=-1;floath=-1|name=tree;caption=Configuration;state=4196348;dir=5;layer=0;row=0;pos=0;prop=100000;bestw=100;besth=100;minw=100;minh=100;maxw=-1;maxh=-1;floatx=-1;floaty=-1;floatw=-1;floath=-1|dock_size(4,0,0)=393|dock_size(5,0,0)=102|'
            perspective2 = None
            position = (50, 50)
            ratio = 2.015267175572519
            showToolbar = True
            size = (1205, 715)
    class CorePluginModule:
        class EventGhost:
            class PythonScript:
                position = (10, 10)
                size = (600, 420)
        class OSE:
            lastFolder = None
        class RadioSure:
            plcmnt = None
            pos = None
        class System:
            class Registry:
                lastKeySelected = 2147483649L
                lastSubkeySelected = 'Software'
                lastValueNameSelected = None
        class Test:
            newval = 100
            oldval = 50
        class Window:
            class FindWindow:
                hideOnDrag = True
```
After starting with EG with -debug and directly exiting EG, my `config.py` now looks like this:
```
autoloadFilePath = u'D:\\topic\\Documents\\Neuer Ordner\\speech.egtree'
checkPreRelease = False
checkUpdate = True
confirmDelete = True
hideOnClose = False
hideOnStartup = False
language = 'en_EN'
lastUpdateCheckDate = (2016, 12, 31)
limitMemory = False
limitMemorySize = 8
propResize = True
refreshEnv = False
showTrayIcon = True
useFixedFont = False
version = 'WIP'
class eg:
    class CorePluginModule:
        class SchedulGhost:
            plcmnt = None
            pos = None
        class SunTracker:
            movingGhost = False
        class Test:
            newval = 100
            oldval = 50
        class Window:
            newval = 100
            oldval = 50
```

The whole `Classes` subtree is gone and a few plugin data (the core ones?). The last class has wrong data (the data from second last?)

kdschlosser:
OK I will revisit the thing. and see what is going on.. 
I do have a question tho.. as this would make this whole process far easier. Should I just institute this for userPluginModule only. since the use of the egplugin files only writes the plugins to the ProgramData folder and forces the plugin to be a userPluginModule.

It will make it far easier to deal with because now I can specifically look at a single class in the Config. instead of having to check the whole thing.

and you know. I do not believe i tested it with the -debug. I had debugging enabled from the GUI. and I just now thought about the order in which the debugging gets set.. we need to do something about having the debug parameter set in such a way that when you select it in the GUI it will set EG to enable debugging right out of the gate. just like specifying the -debug. 

Let me work on a couple of things. the whole damned eg.Log really needs to be re written. it is such a messed up way everything is handled. I do not know if you agree or not. But i do want to poke around at it a bit.
This is going to be easiest if the debugging gets "fixed" and everything gets done in one place
having 3 ways and in 3 different spots in the code that debugging gets set is creating an issue with this..

Try it with just setting the debug from the GUI then restarting EG I bet this is where i made a mistake because I had forgotten how the debugging is done in EG.

I do not want to band-aide the thing to make it work properly because of a problem elsewhere. I think we should really revamp the eg.Log and come up with a really nice simple to follow and understand Log implementation. and have it directly handle the eg.PersistentData instead of burying it in Config which is not a place I would look for code having to do with eg.PersistentData if debugging is set.

I do think we do need to have this in place. But It should be done in a better way. and that is going to require changes. and it lies right in with the discussion I am in about Destroying the frame or not.

I really do not know what direction to head in with this. I think I am going to code up something and see if you like it or not and we can go form there... I do not want to add more confusion to the code. eg's code is already a bloody mystery and has way to much voodoo magic code in it. I don't want to make it worse. Kill this PR and let me come up with something better. If you want maybe start another branch specifically for the logging and debugging and we can shoot things at that instead of beating up the master with PR's that don't get used. and when we come up with something make a clean commit chain and then do a PR to the master.

let me know what ya think.,

topic2k:
I like the idea of removing old data, but i would never do it without asking the user.

If it is easier for you to just look for UserPlugins: do it.

kdschlosser:
I am going to need a branch i am going to make a mess out of...

and this addition does ask the user if they want to remove the data. I would never just do and delete something from someone's computer without asking permission.

and even tho i do ask for permission. it still stores the data until the save button is pressed or the save is click on when prompted if exiting EG. otherwise it will undo it regardless of the fact that they answered yes.

My god could you imaging just going and deleting the data and how many people would scream about it. LOL

BTW. I have rewritten the Config and the PersistentData in such a way that eg.config = eg.Config() is now the first entry in the Core. and setting the debugLevel is second I am working on making the logging the third. 

I set this up in a way that there is no more "Section" class in the Config. it uses eg.PersistentData just like it should because that is really what it is. when i polish it up a little more. I will make a branch on my side and you can take a gander at it. and see what ya think. By doing this is removes the setting of the debugLevel in multiple places but it also moves everything into the persistent data like it should be. and any marking for deletion or marking that it's loaded because of debugging is done there. But in the process I have also corrected another issue. if there is a plugin update and the developer added another attribute to the eg.PersistentData subclass the way EG stands now. a traceback would be thrown when the plugin went to access that attribute. Now it will add the missing attribute when the plugin loads. I can set it up so that it would remove any attributes that have been removed by the plugin developer as well. It is kind of a toss up as to whether that should be done or not. again maybe adding another dialog to ask the user if they want to remove them. I also changed how the config is written to the file. it's not iterated through solely by eg.Config. eg.Config does it's own thing and if one of the attributes is a base class of PersistentData then it will tell that attribute to save it's own information. this is a good thing because that instance can check it's self to see if it has a config or delete flag and also checks to see if it even has any attributes to write all together. and decides to write to the file or not. it's actually a file like object. that gets passed. and when done it runs the code in the object into a dummy namespace to see if there are any errors. and if so doesn't write the actual file. and outputs the traceback along with the object contents to the debug log.

kdschlosser:
ok. so I think I have it done.

It was a complete rewrite of eg.PersistentData as well as eg.Config.

This whole change allows for 

    eg.config = eg.Config()

which means that debug logging can get set right out of the gate.

I have to think about how I want to do the commits on this. so look for it later today.

kdschlosser:
I have a question. I am chasing an illusive issue. but I do know what it is now.

when appveyor installs and runs EventGhost there is a user prompt that will appear to create a new config file. Now I am sure I could do some hocus posus and during the build create the file. and then it would be there when EG is run for the first time. But I do not feel this is the proper way.

Now is there a way that i can check to see if it is appveyor doing the install?.
I thought about using the actual location for the config file. I am not sure what username is used or where the appdata is located. and i do not want to run tests against it to find some kind of a unique identifier. This would add a bunch of commits for this purpose and I do not want to do that. I have been chasing this issue. EG builds and installs fine for me. 

You know. I just thought of how to do it.

topic2k:
> Now is there a way that i can check to see if it is appveyor doing the install?.

You can check for the environment variable `APPVEYOR`. [See here](https://www.appveyor.com/docs/environment-variables/)

kdschlosser:
ok this is sweet. the APPVEYOR  will help out alot and be the proper way. so i will have this thing make the config file if APPVEYOR is an environment variable

kdschlosser:
I have been running this for a couple of days now.. been adding a removing plugins debugging mode, not in debugging mode. running it standing on my head... I haven't had a single issue. You should take a poke at it and see if you can find any problems

eg/Classes/Text.py
```diff
@@ -24,6 +24,8 @@
 
 class Default:
     class General:
+        deleteConfigMessage = "Do you want to remove all configuration data?"
```
topic2k: If i first got that dialog i was scared "what? delete all my config?! why?!". Then i realized it was just for the plugin. So it would be better to mention that:
`deleteConfigMessage = "Do you want to remove all configuration data for plugin '%s'?"`

eg/Classes/PluginItem.py
```diff
@@ -78,6 +78,16 @@ def SearchFunc(obj):
             return False
         if not TreeItem.AskDelete(self):
             return False
+
+        if self.info.persistentData.config:
+            answer = eg.MessageBox(
+                eg.text.General.deleteConfigMessage,
```
topic2k: If i first got that dialog i was scared "what? delete all my config?! why?!". Then i realized it was just for the plugin. So it would be better to mention that:
`eg.text.General.deleteConfigMessage % self.info.name,`

kdschlosser: I think in my original version i did put in plugin configuration data... but somewhere along the line.. it got misplaced? LOL

kdschlosser: changes made

PR #141, opened 2016-12-27
[EventGhost] - Enhancement - MainFrame. On application exit closes open dialogs if no unsaved data

kdschlosser:
When closing EventGhost instead of just having it blink at you if there is a dialog still open this will list all of the open dialogs and ask if you want to close them.

This solves a large inconvenience when creating a plugin and debugging a configuration panel that throws a traceback and offers no way to close it resulting in the developer having to end the process.

### DIFF
--- a/eg/Classes/MainFrame/__init__.py
+++ b/eg/Classes/MainFrame/__init__.py
@@ -580,20 +580,19 @@ class MainFrame(wx.Frame):
         elif len(self.openDialogs) == 0:
             eg.app.Exit()
         else:
+            for dialog in self.openDialogs[:]:
                 panel = getattr(dialog, 'panel', None)
-                    if not panel.isDirty:
                 try:
+                    if not panel.isDirty:
                         dialog.Destroy()
                         continue
-                    pass
                 except AttributeError:
-                self.Iconize(False)
-            for dialog in self.openDialogs[:]:
+                    pass
 
+                self.Iconize(False)
                 dialog.Raise()
                 self.RequestUserAttention()
                 return
-
             self.OnClose(None)
 
     @eg.LogIt
--- a/eg/Classes/MainFrame/__init__.py
+++ b/eg/Classes/MainFrame/__init__.py
@@ -545,29 +545,20 @@ class MainFrame(wx.Frame):
             else:
                 eg.app.Exit()
         else:
-            dialogNames = []
-            for dialog in self.openDialogs:
-                dialogName = dialog.GetTitle() + ' - '
-                if hasattr(dialog, 'headerBox'):
-                    dialogName += dialog.headerBox.GetChildren()[0].GetLabel()
-                else:
-                    dialogName = dialogName[:-3]
-                dialogNames.append(dialogName)
-            message = (
-                eg.text.General.closeOpenDialogMessage +
-                '\n'.join(dialogNames)
-            )
-            answer = eg.MessageBox(
-                message,
-                eg.text.General.closeOpenDialogCaption,
-                style=wx.YES_NO | wx.ICON_QUESTION
-            )
-            if answer == wx.ID_YES:
-                for dialog in self.openDialogs[:]:
-                    dialog.Destroy()
-                self.OnClose(None)
-            else:
+            for dialog in self.openDialogs[:]:
+                panel = getattr(dialog, 'panel', None)
+                try:
+                    if not panel.isDirty:
+                        dialog.Destroy()
+                        continue
+                except AttributeError:
+                    pass
+
+                dialog.Raise()
                 self.RequestUserAttention()
+                return
+
+            self.OnClose(None)
 
     @eg.LogIt
     def OnDialogDestroy(self, event):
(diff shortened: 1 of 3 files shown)


topic2k:
Wouldn't the expected behaviour be to hear a bell and flash the window? Don't know if there is style guide for this. Would like to hear what others say.

kdschlosser:
I understand what the "standard is for this. and I can have it still do the bell and flash the title bar of EG but also show the dialog.

the big reason for this is the huge annoyance of...
If you are coding up a dialog. for an action or plugin.. and something goes wrong and a traceback is thrown. you are unable to close the dialog because it was never shown and EG will not let you exit because it thinks there is a dialog open. Now that i have spent more time with the core code. i believe i can fix that issue without the need for this PR. but it's also a nice convenience for the user.

I would still like to fix the traceback issue. and will do so in the next couple of days and shoot over a PR for that.. up to you or the users on if they want to have a dialog for this.

ThomasBott:
I'm in for the fix of the trackback issue, but in general a dialog telling me a dialog is open is a bit of redundant, I can see if something is open - especially when it blinks ;-)

kdschlosser:
It might be.. but this was intended to be the fix for the "misplaced" dialog problem... but now that i have become a little more knowledgeable about the core code. I am pretty sure I know where the issue is and how to take care of it. I am going to attempt a direct fix tonight.. if successful I will issue a pull request for it. and if not. I will keep on poking at it with a stick until i find where the problem lies

kdschlosser:
@ThomasBott 
I just did a commit with some modifications.
I removed the message dialog because it would be obsolete to the changes listed below. 
It now checks the open dialog to see if it is a standard EG dialog. If so it will check to see if there is any data that needs to be saved.

If it is not a standard EG dialog or there is unsaved data it will raise the dialog and do the blinky blinky beep thing. Otherwise it will close the dialog.

ThomasBott:
@kdschlosser Sorry for the late reply, I must have missed your last comment.

Awesome! Works for me :)
Thank you!

One thing that we may consider is adding the same functionality to eg.app.Restart(), as currently it just closes all dialogues without checking (it was always like that)

kdschlosser:
I will issue another PR for that it shouldn't be an issue. 
But i do have a question on that.. should we modify the eg.app.Restart or create a new function for this. the reason for asking is I think we should maybe consider adding some parameters one that will save EG before closing and the other would be to close open dialogs.

the process would work like this... 

autoSave = True, closeDialogs = True
when closing any open dialogs. if there is unsaved work. it will click on the OK button then it will close the dialogs.. and save EG and restart

autoSave = False, closeDialogs = True
will close open dialogs. if one of them has unsaved work. it will stop right there. otherwise it will close the dialogs. if the tree is dirty then it will stop right there. otherwise it will close.

I am sure you got the other 2 ways. but at least this way there are some choices so if you did want to use this programmatically. you have some choices on what to do. 

I just don't know if it should be added to eg.app.Restart() or a new function under Utils.





ed. then it will save EG and restart it. if the closing of the dialogs is false and the save is true.. if there are any open dialogs it will not do a thing at that point..
 


I think we should leave eg.app.Restart the way it is.. and create a new function

    eg.Utils.RestartEventGhost(save=False, closeDialogs=True)

ThomasBott:
To be consistent, I'd treat eg.app.Restart() similar to eg.app.Exit(). So I'd not modify the eg.app.Restart() function itself, but the one triggered by the UI action - and if there is none yet, create one.

kdschlosser:
I am going to do the work on this tomorrow in the AM. it should be fairly easy to do. that way we can get this thing done and added.

kdschlosser:
I wanted to give you an update to let ya know that I have been working on this.. bit it is proving rather hard. because of the saving of data in an open dialog and the use of the tasklets what is happening is if there is data to be saved.. it sends the command to "click" the ok button.. but because that command is send off to a different thread the function gets to continue to run and it will end up passing right by the saving of the egtree because the tasklet is still processing and has not set the data yet for the OK button press. and EG doesn't get told that it needs to be saved before the function gets to the point in checking. so the function think everything is dandy and nothing needs to be save when in reality it does causing nothing to get saved. and the use of while loop to check the EG save state to wait for it to get set always causes EG to crash.

kdschlosser:
OK i believe I have this all sorted out now.. I had to do a merge because it has been kicking around for a while. and things that related to it had been changed. The restart menu Item has been changed and I have added eg.Utils.Restart(bool(save), bool(closeDialogs))

kdschlosser:
Well the poll in the forum has been up for a while and it's unanimous in favor of this. I personally think it should be in eg 0.5. 
I think the users should be able to enjoy this sooner then later

kdschlosser:
Knock knock. House keeping..

topic2k:
"Come in, door is open" 😄  (as we are in RC mode, i think that's something for 0.5.1)

kdschlosser:
well fantastic. i do not know if i can create a 0.5.1 milestone i will try

PR #142, opened 2016-12-28
Change how the MainFrame handles Id's

kdschlosser:
Replaced use of collections.defaultdict with a custom class that checks wx to see if the attribute being asked for exists. If it does gets the attribute from wx and sets it as an attribute of the class ID. This also allows for getting ID's in a dictionary fashion ID['ID_NAME'] or ID.ID_NAME.
Removed collections.defaultdict import
Re-pointed all  wx.ID_ items to ID.

### DIFF
--- a/eg/Classes/MainFrame/__init__.py
+++ b/eg/Classes/MainFrame/__init__.py
@@ -40,23 +40,21 @@ ADD_MACRO_ICON = CreateBitmapOnTopOfIcon(ADD_ICON, eg.Icons.MACRO_ICON)
 ADD_EVENT_ICON = CreateBitmapOnTopOfIcon(ADD_ICON, eg.Icons.EVENT_ICON)
 ADD_ACTION_ICON = CreateBitmapOnTopOfIcon(ADD_ICON, eg.Icons.ACTION_ICON)
 
-ID_DISABLED = wx.NewId()
-ID_EXECUTE = wx.NewId()
-ID_PYTHON = wx.NewId()
-ID_TOOLBAR_EXECUTE = wx.NewId()
+class ID(dict):
+    def __getitem__(self, item):
+        return getattr(self, item.upper())
 
-ID = defaultdict(wx.NewId, {
-    "Save": wx.ID_SAVE,
-    "Undo": wx.ID_UNDO,
-    "Redo": wx.ID_REDO,
-    "Cut": wx.ID_CUT,
-    "Copy": wx.ID_COPY,
-    "Python": ID_PYTHON,
-    "Paste": wx.ID_PASTE,
-    "Delete": wx.ID_DELETE,
-    "Disabled": ID_DISABLED,
-    "Execute": ID_EXECUTE,
-})
+    def __getattr__(self, item):
+        item = item.upper()
+        if hasattr(wx, 'ID_' + item.upper()):
+            attr = getattr(wx, 'ID_' + item.upper())
+        else:
+            attr = wx.NewId()
+        setattr(self, item, attr)
+        return attr
+
+ID = ID()
+
 
 Text = eg.text.MainFrame
 
@@ -396,12 +394,12 @@ class MainFrame(wx.Frame):
         # the menu command OnCmdExecute will be used in conjunction to
         # our special mouse click handlers
         toolBar.AddSimpleTool(
-            ID_TOOLBAR_EXECUTE,
+            ID.TOOLBAR_EXECUTE,
             GetInternalBitmap("Execute"),
             getattr(text, "Execute")
         )
 
-        toolBar.EnableTool(wx.ID_SAVE, self.document.isDirty)
+        toolBar.EnableTool(ID.SAVE, self.document.isDirty)
         toolBar.Realize()
         self.SetToolBar(toolBar)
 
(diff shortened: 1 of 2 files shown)


topic2k:
Where will this change be usefull (other then in MainFrame/__init__)?

kdschlosser:
That is kind of what i was thinking about is making the ID a global thing... turn it into a "Core" item so it could be used to manage all wx ID's a type of convenience class so if there si a WX one named that way then it will use it. but it also got rid of that block of code in my IDE that was always marked as being incorrect because of passing a dict literal into the constructor for defaultdict.

But this does have greater usefulness. to be able to actually have a widget trigger an event and have that event show up in some other class that has no prior knowledge of the widget but has the ability to simply check to see if the ID's match.

topic2k:
Sorry, still don't get it.   :confused:  Maybe a code example would help.

> but it also got rid of that block of code in my IDE that was always marked as being incorrect because of passing a dict literal into the constructor for defaultdict.

Don't know if that would justify the code changes  :relaxed:  maybe use another IDE instead :interrobang:    :grinning:

kdschlosser:
I am aware. of the IDE thing. But that is what made me look at it. and I said this can be done better then having to import another module to do the same task that very few lines of code could accomplish...

Now I know that the import of collections would have a very minimal impact on performance and memory etc. but if you add up all of those impacts throughout the use of EG they do add up.

this can also be turned into a convince class for all of EG. instead of being limited to just the mainframe.. call it a proof of concept.. If not no worries. 

BTW I am using pyCharm as my IDE and it does a pretty fare shake at pep8 code compliance and errors. not saying it's perfect.

topic2k:
> BTW I am using pyCharm as my IDE and it does a pretty fare shake at pep8 code compliance and errors.

I also use PyCharm 😄 

Before merging any new functions, classes etc. i would like to fix all currently known bugs for a beta 6 release. After that i think it's time to make it a full release.

kdschlosser:
I am just tossing ideas out there and if it's a good one and gets added whoray! and of not. no biggie. But at the very least i always learn something new in the process. and am able to use the knowledge somewhere else

topic2k:
Ok, i think i will merge it, but before i do that, one thing: What about putting the ID class in its own file so it could be used elsewhere in the source?

kdschlosser:
agreed.

I would like to start setting up things in a manner which can be followed.. 

as with the name eg.PluginManager..

any modules that controls or monitors other classes I feel should follow the "Manager" name scheme.

I do not know how you feel about this.. I think it will lend to makig the code easier to follow.
so eg.IDManager.

like the eg.GUID should be eg.GUIDManager and we should consider having these different classes initialize themselves in their module and set the class instance into sys.modules

    class IDManager(object):
        def __init__(self):
            self.__originalmodule__ = sys.modules[__name__]
            sys.modules[__name__] = self
            sys.modules[__name__ + '.' + __name__.split('.')[-1]] = self

    IDManager = IDManager()


by having a setup like above will make make any import or use of it always be the same instance
but it also provides a mechanism to access anything that may be at the module level if needed by using the __originalmodule__ method. where as the current "lazy" import setup does not allow for this.

    from eg.Classes.IDManager import IDManager

    from eg.Classes import IDManager
    
    import eg.Classes.IDManager

    import eg.Classes.IDManager.IDManager
    
    eg.IDManager
    
this is for future discussion but wanted to throw it out there for possibly v1.0? who knows. but I need to come up with some kind of a name for this that we will be able to identify easily. like 
eg.wxID i think is pretty suiting... open to suggestions here

topic2k:
You have great ideas and you know how to it, i like that. So my suggestion is: use this PR to make any needed changes for this and when we are satisfied with it, we will merge it, no matter if it will be 0.5.x, 0.6.x or 1.x.x. Of course we should change the title to something like "change how eg handle it's modules".

And if you want me to test something, drop me note. Then i will checkout from your repo and/or download an artifact from appveyor.

BTW: Do you have slack, gitter, wire or another messenger? Sometimes i find it easer to use a messenger then using the issues conversations or the forum.

kdschlosser:
this PR is specifically a lazy wxID import mechanism..

so if there is a wxID already in existence with the name provided it will use it. and if there isn't it will create a wx.NewId(). it's a nicety. but will also allow for capturing events from other dialogs without the need to access the specific widget you would just bind to the dialog for the specific event and ID. because the ID's would now be able to be globally accessed. it gives the availability to reduce class level attributes being set for widgets because this occupies more memory when a class level attribute for a widget is set. even after destruction there is still a holder for a pyDeadObject. so by using the ID's you can do a class level definition if you want for an event. but now have an easy means to identify the widget. and since you don't have to set the class level attribute for the widget the widget will get deleted when the method closes

That is my understanding of how this can benefit. tell me if i am wrong

kdschlosser:
I don't have any messengers installed. so tell me what one you prefer and i will go ahead and install it

topic2k:
you don't need to install, but if you really want, then i would suggest slack. It's connected to GitHub and there will be messages with commits and comments for this repo.

eg-slack

kdschlosser:
it's telling me that you have to invite me..

    Don't have an account on this team yet? Contact the team administrator for an invitation 

I am using the EventGhost.slack domain... don't know if i need to set up my own. but i think in order for the chat bits to work i have to login to the EventGhost domain.

topic2k:
Ah yes, i need a eMail to invite you. kdschlosser@users.noreply.github.com doesn't work ;)

kdschlosser:
I will shoot you a message in the forum with my e-mail address....

Issue #139, opened 2016-12-26
EventGhost doesn't minimize to the task bar


ThomasBott:
EG 0.5 beta5
if ´Display EventGhost icon in system tray´ is disabled, EventGhost doesn't minimize to the task bar (if I'm not mistaking, it did this in version 0.4) I think all 0.5 versions are affected.
You can only bring back the UI by executing the EG .exe

kdschlosser:
I confirmed this behavior in beta 4

ThomasBott:
Replacing the function OnIconize in eg\Classes\MainFrame\__init__.py with this one:

    def OnIconize(self, event):
        """
        Handle wx.EVT_ICONIZE
        """
        # On iconizing, we actually destroy the frame completely
        if eg.config.showTrayIcon:
            self.document.HideFrame()
        else:
            event.Skip()

And function OnClose in eg\Classes\MainFrame\__init__.py with this one:

    def OnClose(self, dummyEvent):
        """
        Handle wx.EVT_CLOSE
        """
        if len(self.openDialogs) == 0:
            if eg.config.hideOnClose:
                if eg.config.showTrayIcon:
                    self.document.HideFrame()
                else:
                    self.Iconize()
            else:
                eg.app.Exit()
        else:
            self.RequestUserAttention()

Resolves this issue, credit goes to @kdschlosser

topic2k:
Confirmed. Please do you or @kdschlosser a pull request.

topic2k:
fixed in #145

PR #145, opened 2016-12-28
Not being able to restore from a minimized state if tray icon is not shown and empty log

kdschlosser:
Fixes the taskbar icon removal when show tray icon is not selected.

Fixes the empty log when maxamizing the frame

EG was set to destroy the frame when minimized. this would cause both of the bugs. because the frame would be destryoed so would the logCtrl, therefore deleting any data that may have been in the log. It would also remove the icon from the taskbar because the frame no longer exists.

I Have changed it to hide the frame if the tray icon is to be displayed and to let windows do it's windows thing if not. This doesn't destroy the logCtrl keeping the log data in tact

### DIFF
--- a/eg/Classes/Document.py
+++ b/eg/Classes/Document.py
@@ -300,8 +300,8 @@ class Document(object):
         if self.reentrantLock.acquire(False):
             if self.frame is not None:
                 if len(self.frame.openDialogs) == 0:
-                    self.frame.Destroy()
-                    self.frame = None
+                    self.frame.Hide()
+                    # self.frame = None
             self.reentrantLock.release()
         else:
             wx.CallLater(100, self.HideFrame)
--- a/eg/Classes/Document.py
+++ b/eg/Classes/Document.py
@@ -19,7 +19,6 @@
 import os
 import wx
 from tempfile import mkstemp
-from threading import Lock
 from types import ClassType
 from xml.etree import cElementTree as ElementTree
 
@@ -63,7 +62,6 @@ class Document(object):
         self.root = None
         self.firstVisibleItem = None
         self.frame = None
-        self.reentrantLock = Lock()
         self.expandedNodes = set()
 
     def AfterLoad(self):
@@ -289,21 +287,6 @@ class Document(object):
             filename = os.path.splitext(os.path.basename(self.filePath))[0]
         return "EventGhost %s - %s" % (eg.Version.string, filename)
 
-    @eg.LogItWithReturn
-    def HideFrame(self):
-        # NOTICE:
-        # If the program is started through a shortcut with "minimize" option
-        # set, we get an iconize event while ShowFrame() is executing.
-        # Therefore we have to use this CallLater workaround.
-        # TODO: Find a better way. Preferable detect the minimize option
-        #       before we create the MainFrame.
-        if self.reentrantLock.acquire(False):
-            if len(self.frame.openDialogs) == 0:
-                self.frame.Hide()
-            self.reentrantLock.release()
-        else:
-            wx.CallLater(100, self.HideFrame)
-
     @eg.LogIt
     def Load(self, filePath):
         if filePath is None:
@@ -358,7 +341,7 @@ class Document(object):
         eg.AsTasklet(eg.UndoHandler.Configure(self).Do)(node)
 
     def Open(self, filePath=None):
-        self.ShowFrame()
+        eg.mainFrame.Iconize(False)
         if filePath is not None:
             res = wx.MessageBox(
                 "Do you really want to load the tree file:\n%s" % filePath,
@@ -456,15 +439,6 @@ class Document(object):
     def SetUndoState(self, undoState):
         self.undoState = undoState
         eg.Notify("UndoChange", undoState)
-
-    @eg.LogItWithReturn
-    def ShowFrame(self):
-        if self.reentrantLock.acquire(False):
-            if self.frame is None:
-                self.frame = eg.MainFrame(self)
-                self.frame.Show()
-            self.frame.Raise()
-            self.reentrantLock.release()
 
     def StartSession(self, filePath):
         eg.eventThread.Func(eg.eventThread.StopSession)()
--- a/eg/Classes/Document.py
+++ b/eg/Classes/Document.py
@@ -298,10 +298,8 @@ class Document(object):
         # TODO: Find a better way. Preferable detect the minimize option
         #       before we create the MainFrame.
         if self.reentrantLock.acquire(False):
-            if self.frame is not None:
-                if len(self.frame.openDialogs) == 0:
-                    self.frame.Hide()
-                    # self.frame = None
+            if len(self.frame.openDialogs) == 0:
+                self.frame.Hide()
             self.reentrantLock.release()
         else:
             wx.CallLater(100, self.HideFrame)
--- a/eg/Classes/Document.py
+++ b/eg/Classes/Document.py
@@ -18,6 +18,7 @@
 
 import os
 import wx
+from threading import Lock
 from tempfile import mkstemp
 from types import ClassType
 from xml.etree import cElementTree as ElementTree
@@ -61,8 +62,36 @@ class Document(object):
         self.filePath = None
         self.root = None
         self.firstVisibleItem = None
+        self.reentrantLock = Lock()
         self.frame = None
         self.expandedNodes = set()
+
+    @eg.LogItWithReturn
+    def ShowFrame(self):
+        if self.reentrantLock.acquire(False):
+            if self.frame is None:
+                self.frame = eg.mainFrame = eg.MainFrame(self)
+                self.frame.Show()
+            self.frame.Raise()
+            self.reentrantLock.release()
+
+    @eg.LogItWithReturn
+    def HideFrame(self):
+        # NOTICE:
+        # If the program is started through a shortcut with "minimize" option
+        # set, we get an iconize event while ShowFrame() is executing.
+        # Therefore we have to use this CallLater workaround.
+        # TODO: Find a better way. Preferable detect the minimize option
+        #       before we create the MainFrame.
+        if self.reentrantLock.acquire(False):
+            if self.frame is not None:
+                if len(self.frame.openDialogs) == 0:
+                    self.frame.Destroy()
+                    self.frame = None
+                    eg.mainFrame = None
+            self.reentrantLock.release()
+        else:
+            wx.CallLater(100, self.HideFrame)
 
     def AfterLoad(self):
         if (
--- a/eg/Classes/Document.py
+++ b/eg/Classes/Document.py
@@ -18,8 +18,8 @@
 
 import os
 import wx
-from threading import Lock
 from tempfile import mkstemp
+from threading import Lock
 from types import ClassType
 from xml.etree import cElementTree as ElementTree
 
@@ -62,36 +62,9 @@ class Document(object):
         self.filePath = None
         self.root = None
         self.firstVisibleItem = None
-        self.reentrantLock = Lock()
         self.frame = None
+        self.reentrantLock = Lock()
         self.expandedNodes = set()
-
-    @eg.LogItWithReturn
-    def ShowFrame(self):
-        if self.reentrantLock.acquire(False):
-            if self.frame is None:
-                self.frame = eg.mainFrame = eg.MainFrame(self)
-                self.frame.Show()
-            self.frame.Raise()
-            self.reentrantLock.release()
-
-    @eg.LogItWithReturn
-    def HideFrame(self):
-        # NOTICE:
-        # If the program is started through a shortcut with "minimize" option
-        # set, we get an iconize event while ShowFrame() is executing.
-        # Therefore we have to use this CallLater workaround.
-        # TODO: Find a better way. Preferable detect the minimize option
-        #       before we create the MainFrame.
-        if self.reentrantLock.acquire(False):
-            if self.frame is not None:
-                if len(self.frame.openDialogs) == 0:
-                    self.frame.Destroy()
-                    self.frame = None
-                    eg.mainFrame = None
-            self.reentrantLock.release()
-        else:
-            wx.CallLater(100, self.HideFrame)
 
     def AfterLoad(self):
         if (
@@ -316,6 +289,23 @@ class Document(object):
             filename = os.path.splitext(os.path.basename(self.filePath))[0]
         return "EventGhost %s - %s" % (eg.Version.string, filename)
 
+    @eg.LogItWithReturn
+    def HideFrame(self):
+        # NOTICE:
+        # If the program is started through a shortcut with "minimize" option
+        # set, we get an iconize event while ShowFrame() is executing.
+        # Therefore we have to use this CallLater workaround.
+        # TODO: Find a better way. Preferable detect the minimize option
+        #       before we create the MainFrame.
+        if self.reentrantLock.acquire(False):
+            if self.frame is not None:
+                if len(self.frame.openDialogs) == 0:
+                    self.frame.Destroy()
+                    self.frame = eg.mainFrame = None
+            self.reentrantLock.release()
+        else:
+            wx.CallLater(100, self.HideFrame)
+
     @eg.LogIt
     def Load(self, filePath):
         if filePath is None:
@@ -370,7 +360,7 @@ class Document(object):
         eg.AsTasklet(eg.UndoHandler.Configure(self).Do)(node)
 
     def Open(self, filePath=None):
-        eg.mainFrame.Iconize(False)
+        self.ShowFrame()
         if filePath is not None:
             res = wx.MessageBox(
                 "Do you really want to load the tree file:\n%s" % filePath,
@@ -468,6 +458,16 @@ class Document(object):
     def SetUndoState(self, undoState):
         self.undoState = undoState
         eg.Notify("UndoChange", undoState)
+
+    @eg.LogItWithReturn
+    def ShowFrame(self):
+        if self.reentrantLock.acquire(False):
+            if self.frame is None:
+                self.frame = eg.mainFrame = eg.MainFrame(self)
+                self.frame.Show()
+            else:
+                self.frame.Iconize(False)
+            self.reentrantLock.release()
 
     def StartSession(self, filePath):
         eg.eventThread.Func(eg.eventThread.StopSession)()
--- a/eg/Classes/MainFrame/__init__.py
+++ b/eg/Classes/MainFrame/__init__.py
@@ -541,7 +541,11 @@ class MainFrame(wx.Frame):
         """
         if len(self.openDialogs) == 0:
             if eg.config.hideOnClose:
-                self.document.HideFrame()
+                if eg.config.showTrayIcon:
+                    self.document.HideFrame()
+                else:
+                    self.Iconize()
+
             else:
                 eg.app.Exit()
         else:
@@ -590,7 +594,6 @@ class MainFrame(wx.Frame):
         """
         Handle wx.EVT_ICONIZE
         """
-        # On iconizing, we actually destroy the frame completely
         if eg.config.showTrayIcon:
             self.document.HideFrame()
         else:
--- a/eg/Classes/MainFrame/__init__.py
+++ b/eg/Classes/MainFrame/__init__.py
@@ -227,10 +227,11 @@ class MainFrame(wx.Frame):
         self.SetAcceleratorTable(self.acceleratorTable)
         self.logCtrl.Bind(wx.EVT_SIZE, self.OnLogCtrlSize)
 
-        if eg.config.hideOnStartup or eg.startupArguments.hideOnStartup:
-            self.Iconize(True)
-        else:
+        if not(eg.config.hideOnStartup or eg.startupArguments.hideOnStartup):
+            self.Show()
             eg.EnsureVisible(self)
+        else:
+            self.Iconize(True)
 
     if eg.debugLevel:
         @eg.LogIt
@@ -603,7 +604,6 @@ class MainFrame(wx.Frame):
 
     def OnIconize(self, event):
         self.Iconize(event.Iconized())
-
 
     def OnLogCtrlSize(self, evt):
         if self.mainSizeFlag:
--- a/eg/Classes/MainFrame/__init__.py
+++ b/eg/Classes/MainFrame/__init__.py
@@ -126,6 +126,7 @@ class MainFrame(wx.Frame):
         iconBundle.AddIcon(icon)
         self.SetIcons(iconBundle)
 
+        self.Bind(wx.EVT_ICONIZE, self.OnIconize)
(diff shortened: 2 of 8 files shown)


ThomasBott:
Thx @kdschlosser !
Seems to solve the two issues!
But created another one: The option "Minimize on close" doesn't work anymore...

kdschlosser:
ok see that is why we test...

not a biggie I will change up one other thing and that should fix it..

gimme a few

also let me know if you have any issues with the starting hidden and anything of that nature. i forgot those options were available.

ThomasBott:
...looks like the new trouble comes from the change to Document.py

I'll let you know if I find more issues ;-)

kdschlosser:
the minimize on close should be good now. give it a test

kdschlosser:
I never did understand why the frame would get destroyed when you minimized EG. it has been like that for as long as i can remember. that's just like the 8mb of memory thing. i think that can really be depreciated. with the way computers are now i don't think that limiting eg to 8mb is going to make an impact on memory consumption that would change anything. WAYYYYY back when EG was made ram was a commodity many PC's didn't have. so I can see where destroying the UI when minimized would help. I don't think that having to do that is really needed anymore. most computers these days come with 8 gig. but at the very least 4. not 1. and ram is so cheap nowadays. hell i can remember spending 300.00 USD for 1 meg. and a 212 meg Western Digital HDD was 400.00 USD

kdschlosser:
and the trouble actually came in the document and the mainframe __init__.

me being the dingdong that i am didn't remove the checking if the frame is None. well duh! of course it's not None I'm not destroying it anymore.

I really should remove the frame hiding and showing from document, as it really has nothing to do with document.

I am going to modify the code

ThomasBott:
> EG was set to destroy the frame when minimized. 

I'm not sure if that's the right way to go, this was also the case in the past, but it never caused the Log to be discarded, we should find the root cause of the issue before changing this part.

kdschlosser:
I know what caused it. found it when i was working on this. it's because of the debugging changes that were made. to have it show in the log and write to the file as well. there is a cache that is supposed to populate when the frame is closed. and it doesn't do this properly. due to the mechanics of how the logging section is written. now this cache does get used when EG first starts.

it was by far a lot easier to modify EG to actually work the way it should. and not this crazy thing of destroying the frame when you minimize it and cache any log entries.
having the frame not be destroyed makes it a whole lot easier to add anything to the EG UI if needed because the frame always exists.. but this also allows for things like what i did with this current commit. and that is minimizing it if you have any dialogs open instead of how it was. which was to disable the minimize button. I think that all in all this functions how it should be. and also functions in completeness. all of the missing bits due to the destruction of the frame are no longer missing.

give this whirly bird a twirl and see if it's proper to you. I do know the debugging end of things needs an overhaul. I have already made up a pane to have the UI show for the debugging with a tree in it for different dates. with specific labels for tracebacks, calls, informational. things of that nature. and it all works. I want to add the ability to do searches. then i will issue a pull request on it. it will be nice to not have debugging flood the log. I have already added the ability to clear the log from inside of EG even tho the file is in use. I already got that all worked out,, :-D

ThomasBott:
> it was by far a lot easier to modify EG to actually work the way it should.

I think this depends on who you ask ;-) 
It must have been made for some good reason, everybody knows that it's easier to hide a frame than to destroy and recreate it.
If the reason was to save 5MB of RAM, I agree that this is not worth the extra effort, but if it positively impacts performance - even if it's not much - we should keep/recover it, in my opinion.

kdschlosser:
hmmm... it still has to go through the normal process...

and there are separate threads for actions, events and the UI (main thread)  so i couldn't see how the UI would have any performance impact. it doesn't run on the same "highway"

would have to check it out and see but even if you are talking a 1 millisecond difference and that's a lot each action. you would have to have 1000 actions fire in succession to make a 1 second impact.

I believe this was by original design to limit the amount of used memory by EG. you have to remember it was made during the XP days. and with the minimum requirements for XP being 64 meg 128 recommended. i can see where even 5 meg would make a difference. LOL

kdschlosser:
and also 5 meg of an 8 meg footprint (if set) is a whole lot. if you only have 64 meg of memory. and why have an setting to limit to 8 meg??? this is why. I'm telling ya. and if there was a performance difference back then it would be so negligible now.

and in all honesty any XP support should be dropped. and anything XP related in the code should be as well.
windows vista isn't to far behind. LOL if it can run vista it can run windows 7. so there is really no purpose to having dedicated code for it. My opinion. But not my choice.

ThomasBott:
OK, I also did some measures on my system (1.5GHz Celeron x2), I assigned EG to run on one core only.
Used my normal EG config triggering 1000 events while minimized EG to tray

Destroy Frame (before the change):
3.56900000572
3.88899993896
3.85700011253
3.89800000191

Hide Frame (after the change): 
3.78099989891
3.81799983978
3.72300004959
3.8180000782

as you see I ran multiple tests and the results were similar. It indeed doesn't look like it makes much of a difference destroying or hiding the Frame, the small differences you see here may have some other reason.
Interesting is, that I can see a much higher CPU load in the task manager, caused by EG, when the frame is hidden and not destroyed! (about 20% higher = 40% higher for one core, reproducible)

So I have some mixed feelings about this change, it seems to eat a lot more recourse, at some point, but without an impact on the Event processing speed. 

btw. I also noticed that EG always starts hidden now, even if it wasn't before (e.g. if you click file -> restart). The other things seem to work though

kdschlosser:
oh poop i did push the last change. i added Show() sorry about that thought i did

I do have a ? for ya.

how old is that box you are running EG on?

kdschlosser:
ans yes it will keep using the resources even if minimized, I don't know of any applications that release their resources when minimized. we really would have to look tho. because i do believe from the constant destroying and recreation could also not allow you to recapture the resources. I would have to come up with a way to programmatically open and minimize EG a bunch of times

kdschlosser:
Destroy Frame (before the change):
3.56900000572
3.88899993896
3.85700011253
3.89800000191

Hide Frame (after the change):
3.78099989891
3.81799983978
3.72300004959
3.8180000782

is this correct????


because it's faster to not Destroy the frame. Again it's such a small amount but it could have something to do with the buffering process.

Old, sum([3.56900000572, 3.88899993896, 3.85700011253, 3.89800000191]) / 4 = 3.80325001478
New, sum([3.78099989891, 3.81799983978, 3.72300004959, 3.8180000782]) / 4 = 3.78499996662

kdschlosser:
you will have to have stackless python 2.7.9 installed and psutil installed into that python in order for the tests to work in eg 0.5

i have some code you can run. send me a message in the forum and i will send it to you. but what it does is it grabs every single little piece of system data. then it simulates opening and closing EG 200 times. 

and guess what the memory consumption in specific is almost exactly identical

because EG 0.5 is of a different python version and different wx version it does consume more memory out of the box. but that is not the issue

in EG 0.4 memory consumption goes up.

I have 8gb ram in this machine
Bone stock versions of EG except the destruction of the frame
I am going to run the same 
here are the memory findings
0.4
memory_percent
    before - 0.595414975571
    after  - 0.619550197685

0.5 modified to not Destroy the frame
memory_percent
    before - 0.780531175218
    after  - 0.780674269421

0.5 unmodified
memory_percent
    before - 0.780641081105
    after  - 0.79828936605

Destroying goes up by a decent chunk. So I would say that there is memory fragmentation or a memory leak.

i can zap ya over the tree for running this. you will have to make one change to the script if you are running it on a 0.5 modified and unmodified. otherwise it will sense between 0.4 and 0.5

ThomasBott:
The hardware is maybe 3 years old(max). Running a celeron 887 with 8GB RAM and SSD.
I designed it to be very energy efficient (for 24/7 operations)
I'm sure that a lot of people use hardware for EG that is wors and they will suffer under performance cuts you don't even notice with faster PCs.

I bet if I ran the test more often, the results will get even closer to each other, I'd really say that the event processing speed didn't change at all with my hardware. Anyway, more CPU cycles seem to be wasted now for things that are not related to triggering and processing the event. By 20% I refered to the overall usage of the CPU which increased from 30% to 50% while running that test.

As I said , I'm not concerned about the memory usage - also in my tests the usage of memory was similar to as it was with destroying the frame, but even it it wasn't, I don't mind as long as it's still below 100MB. BTW, the option to limit the RAM usage when minimized is useless indeed ;-)

kdschlosser:
up to you I don't know if i got the forum user correct when i sent over the tree for the testing of the event timings. I was only guessing as to who thomasbot is. but i thought i read somewhere it was the person i sent it to. 

so let me know if it's who i think it is. i will send ya over the tree I used to run these tests

kdschlosser:
hey check this out...

Stackless python does not make use of any kind of multi-core environment it runs on.
This is a common misconception about Stackless, as it allows the programmer to take advantage of thread-based programming. For many people these two are closely intertwined, but are, in fact two separate things.

Internally Stackless uses a round-robin scheduler to schedule every tasklet (micro threads), but no tasklet can be run concurrent with another one. This means that if one tasklet is busy, the others must wait until that tasklet relinquishes control. By default the scheduler will not stop a tasklet and give processor time to another. It is the tasklet's responsibility to schedule itself back in the end of the schedule queue using Stackless.schedule(), or by finishing its calculations.

all tasklets are thus executed in a sequential manner, even when multiplpe cores are available.

The reason why Stackless does not have multi-core support is because this makes threads a whole lot easier. And this is just what stackless is all about

the way windows works is if you set the affinity to one core. and that process maxes out that one core. if you have a 4 core machine and set the affinity to all cores. the use will stop at 25% for each core. equaling 100% for a single core.

so the result you are seeing is no different then setting it to use multiple cores. now here is another hitch.

because of stackless and the way the dialogs are EG by design will never use multiple cores. and having a computer that has multiple cores in no way would improve the speed of EG. because stackless does not allow for multi core processing. in fact it really doesn't even allow for multi thread processing. because of it's round robin way of running the tasks. 

stackless python is in essence slowing down EG.

in order to get true multi core processing one would have to create a separate processes for the GUI, and for the events, and for running the actions..

ThomasBott:
Interesting, I didn't know that about stackless. Then it's even more important to save CPU cycles where we can ;-)

I don't say we need to destroy the frame, but we should make sure EG 0.5 beta X doesn't consume much more resources than 0.5 beta 4. I don't know what causes the additional load, but something does for sure.

kdschlosser:
I do have question. how are you producing your measurements?

everything I have tried thus far has indicated that not destroying the frame is faster. while using less (relative) resources.

to trigger 1000 events I get a speed of 45ms without having the frame destroyed. compared to 90ms with it destroyed, memory use goes up when destroyed. and when it's not it stays about the same. I have triggered events to go through the whole process. I have simulated printing events by going to eg.log.Write and removed TriggerEvent has a factor in checking the speed of putting the data in a cache vs having it go right on thru. 

I have no idea why it is taking 4 seconds for you. there has to be some other factor causing this kind of latency

ThomasBott:
As I wrote, I tested it with my productive egtree, lots of macros, lots of a plugins. It may be a combination of things, but after the change I had a much higher CPU load, checked that multiple times.

Anyway, as the bug is now fixed where it orininaly came from, you may want to save this change for version 0.6?
It's a bigger thing after all and not the kind of change that is applied during the beta phase of an release. Also I guess you did not remove the caching thing yet, witch is useless when the frame isn't destroyed any more, right?

BTW. #139 , do you want to create the PR for that? I mean, it was you who fixed it ;-)

kdschlosser:
oh no no I agree with ya. and there is no way to fix the not being able to open EG back up if you minimize it and do not have the tray icon shown.

but.. this can be included in the current version so it would give the best of both worlds. if you have the tray icon turned on and minimize it would destroy the frame. and if you do not have the icon on it would not destroy it. that is pretty much the only way it would work. unless you always had the icon visible on minimize and the option was to display it when not minimized.

kdschlosser:
OK I reworked this PR to take care of the other half of the issue while keeping the current bugfix for the empty log..

So now it will destroy the frame if the tray icon is shown. and if the icon is not shown it will iconize EG and any any dialogs that may be open.

kdschlosser:
sorry about that. I had made the changed for the trayicon.. i just forgot that i made them and didn't include them in the commit.

The other is fixed as well.. I was going to make this change so it wouldn't always assume to Hide the frame when calling Iconize. I forgot to make the change. but it's done now...

eg/Classes/Document.py
```diff
@@ -62,10 +62,37 @@ def MakeCls(name):
         self.filePath = None
         self.root = None
         self.firstVisibleItem = None
-        self.frame = None
         self.reentrantLock = Lock()
+        self.frame = None
```
topic2k: Same here: Why move the that line?

eg/Classes/Document.py
```diff
@@ -18,8 +18,8 @@
 
 import os
 import wx
-from tempfile import mkstemp
 from threading import Lock
+from tempfile import mkstemp
```
topic2k: Why is the import moved to another line? Does it make any difference? If not, i would like to have the change undone to make the diff shorter.

eg/Classes/Document.py
```diff
@@ -62,10 +62,37 @@ def MakeCls(name):
...
+        if self.reentrantLock.acquire(False):
+            if self.frame is not None:
+                if len(self.frame.openDialogs) == 0:
+                    self.frame.Destroy()
+                    self.frame = None
+                    eg.mainFrame = None
+            self.reentrantLock.release()
+        else:
+            wx.CallLater(100, self.HideFrame)
+
```
topic2k: If the functions were not moved, it would make it easier to see the difference. Look [here](https://github.com/EventGhost/EventGhost/compare/master...topic2k:bugfix-trayicon-emptylog).

eg/Classes/MainFrame/__init__.py
```diff
@@ -126,14 +126,14 @@ def __init__(self, document):
...
 
-        self.Bind(wx.EVT_ICONIZE, self.OnIconize)
         self.Bind(wx.EVT_MENU_OPEN, self.OnMenuOpen)
         self.Bind(wx.EVT_CLOSE, self.OnClose)
         self.mainSizeFlag = True
         self.ratioLock = False
         self.ratio = Config.ratio
         self.Bind(wx.EVT_SIZE, self.OnSize)
         self.Bind(wx.EVT_MOVE, self.OnMove)
+        self.Bind(wx.EVT_ICONIZE, self.OnIconize)
```
topic2k: same here: revert

eg/Classes/MainFrame/__init__.py
```diff
@@ -226,6 +226,7 @@ def OnEnterKey(dummyEvent):
         )
         self.SetAcceleratorTable(self.acceleratorTable)
         self.logCtrl.Bind(wx.EVT_SIZE, self.OnLogCtrlSize)
+
```
topic2k: well, i think thta one is ok 😉

eg/Classes/MainFrame/__init__.py
```diff
@@ -526,7 +527,7 @@ def OnAddDialog(self, dialog):
             return
         self.openDialogs.append(dialog)
         dialog.Bind(wx.EVT_WINDOW_DESTROY, self.OnDialogDestroy)
-        self.SetWindowStyleFlag(~(wx.MINIMIZE_BOX | wx.CLOSE_BOX) & self.style)
+        # self.SetWindowStyleFlag(~(wx.MINIMIZE_BOX | wx.CLOSE_BOX) & self.style)
```
topic2k: why not remove it completly?

eg/Classes/MainFrame/__init__.py
```diff
@@ -554,8 +555,8 @@ def OnDialogDestroy(self, event):
             self.openDialogs.remove(dialog)
         except ValueError:
             pass
-        if len(self.openDialogs) == 0:
-            self.SetWindowStyleFlag(self.style)
+        # if len(self.openDialogs) == 0:
+        #     self.SetWindowStyleFlag(self.style)
```
topic2k: why not remove them completly?

kdschlosser:
You have to remember this pr originally removed the destruction of the frame completely. so things that may have gotten moved around is because they were deleted at one point. and then re added to accommodate for both the destruction of the frame and the iconization. 

I will reshift things.. but you also pointed out to me the frame style changes i made. and it's a good thing i commented that line :P because i would have completely forgotten about. so now i know that i have to make accomidations.. because the style of the frame has to change dynamically with the changing of the show tray icon. so this may take a little bit to work through. I will issue a new PR on this to keep the commit history clean.

topic2k:
If those changes are neede for this bugfix (PR), then commit them here. I will squash all commits to one before merging.

kdschlosser:
ok i can do that. i will finish this up this evening.. clean out the code and take care of the frame style... it will be a done deal this evening

ThomasBott:
#139 
Aren't these the only two things you need to change?
Wouldn't it be much simpler to just create a new PR instead of readding old code you removed in the first place?

kdschlosser:
OK code cleanup is done.
The style changes are put into place.
I also added in the options dialog to set the frame style as well. this is so that the changes become effective right after closing the dialog.

kdschlosser:
that will work but will cause issues when the minimize to tray on close is turned on.

and also the way I have set it up will allow for the minimizing of EG whether or not a dialog is open. if you don't handle the frame style changed the minimize button will be grayed out even if the tray icon is not shown.

ThomasBott:
Well, It's true that you currently can't minimize EG if a dialogue is open (I don't think that's too bad tough)
But I don't see any issue with minimize on close.
Anyway, was just an idea.

kdschlosser:
the minimize on close would have been the same deal with the style thing... 
either or as long as it's fixed... i am all for it.
but  i do like being able to minimize even if i have dialogs open tho :-D

kdschlosser:
the latest commit takes care if creating the mainframe and then iconizing it if the trayicon is not shown and -hide is provided at the CLI

kdschlosser:
@ThomasBott yeah It would be best to have made a new PR. but i knew there would be things i missed and instead of boogering up the commit history in a new PR I figured this one would probably get squashed and I would make a new one anyway. so may as well get the code right first then do another PR if that is what @topic2k wanted

topic2k:
@kdschlosser can you explain why the window flags need to be changed? I just wanna understand what would be else the problem. Oh, and maybe add it as a short comment to the SetWindowStyleFlag function. 😉 
Beside that, your are done with this PR?

kdschlosser:
With the window style flag if the tray icon is snow it can be minimized without a problem. it will also minimize any open dialogs. because there is no destruction of the frame there is no reason why the minimize button should be available and function like it should. now if the tray icon is shown and the destruction of the frame is supposed to take place. that means that any child of the frame also has to be destroyed.. hence why the minimize button would be grayed out. not until all child frames/dialogs are closed will it become available so the frame can be minimized..

I added the last bit of code last night. which was for the hideOnStartup. if you want to go a different route with this we can. @ThomasBott solution is far simpler. the only thing would have to be added to his solution would be the hideOnStartup but then the styles can stay the same. however i do like having the eg.mainFrame and if that could stay would be nice. instead of having to type in eg.document.frame.

topic2k:
Now worries, i just wanted to understand why it's needed. I still wanna merge this PR, just waiting for your "GO"🏁

kdschlosser:
OK let me add that comment... do you think instead of comments it would be better to start doing docstrings instead??? this way an explanation of what it's for and how to use it can be use in coding help files or show up in an IDE?

topic2k:
In that case it fits to a docstring. Would it have been somewhere in the middle of a function and only apply to some code of lines, than i would use comments above that code part.

kdschlosser:
I added the docstring for the SetWindowStyleFlag() so it should be good to go

I would really like to start getting into the habit of doing that for any new PR's. So remind me in case I forget to do it.. We have to do it. So we know how things work. But it will also make our life easier when we are coding to be able to bring up some help right inside the IDE instead of having to open the code for that function/method/class and look it at it to figure out what is going on, and how to use it.. A little time spent now will save a lot of time in the future..

kdschlosser:
and yes agreed on code modification. but a new method/function/class we have to start adding those docstrings. I know it really sucks doing it... and I hate it as well. But think about how easy it can make coding..

eg/Classes/MainFrame/__init__.py
```diff
@@ -521,12 +521,45 @@ def GetPanelDirection(self, panel):
         pos = info.find(";dir=") + 5
         return info[pos]
 
+    def SetWindowStyleFlag(self, *args, **kwargs):
+        """
+        Changes the main frame style depending on the display of the tray icon.
+
+        sets the stlye flags fo the minimize button will be grayed out if the
```
topic2k: typo/change:
 Sets the stlye flags for the minimize button. The button will be grayed out if the

eg/Classes/MainFrame/__init__.py
```diff
@@ -521,12 +521,45 @@ def GetPanelDirection(self, panel):
...
+        tray icon is shown and there are any open child windows.
+        If the tray icon is not shown the minimize button will function as
+        usual.
+
+        :param args: unused parameter
```
topic2k: unused but needed to be compatible with the default SetWindowStyleFlag function, right?

kdschlosser: nope.. removed the ability for anyone to set the style parameter using the method..
this is intentional. don't want an accidental change. a deliberate one can be done using

  wx.Frame.SetWindowStyleFlag(eg.document.frame, style)

if needed. i don't see where it would be but in the event that it does. we will know it wasn't by accident (in our code) if the call above is done.

kdschlosser: so it won't toss out a traceback. I am not sure if the internals of wx use it at all so i wanted to make sure that it wouldn't cause an issue.

eg/Classes/MainFrame/__init__.py
kdschlosser: ooops hehe!. I have bad arthritis. And my fingers get very lazy to that shift key.

eg/Classes/MainFrame/__init__.py
kdschlosser: last commit is a better docsting. and fixes that typo

eg/Classes/MainFrame/__init__.py
```diff
@@ -521,12 +521,47 @@ def GetPanelDirection(self, panel):
         pos = info.find(";dir=") + 5
         return info[pos]
 
+    def SetWindowStyleFlag(self, *args, **kwargs):
+        """
+        Changes the main frame style depending on the display of the tray icon.
+
+        Sets the stlye flags fo the minimize button will be grayed out if the
```
topic2k: I'm sorry to say, but... :worried:
Sets the stlye flags **for** the minimize button **. The button** will be grayed out if the

kdschlosser: well hell I guess there was more then one typo LMAO

it's been one of those days..

kdschlosser:
Not sure if this is good to go or not.. I have made all of the requested changes... Just wondering..

PR #146, opened 2016-12-30
Add checking for hidden or system files on plugin import/export

kdschlosser:
problem:
There could accidentally be thumbs.db or desktop.ini or other hidden or system files/folders that inadvertently get into the plugin folder. These files/folders will get copied into the egplugin file. 

solution:
Added checking for System or hidden attributes for files/folders within the plugin directory. If found does not include them in the plugin package

### DIFF
--- a/eg/Classes/PluginInstall.py
+++ b/eg/Classes/PluginInstall.py
@@ -64,76 +64,6 @@ INFO_FIELDS = [
     "icon",
 ]
 
-UNCHECKED = (
-    '\x89\x50\x4E\x47\x0D\x0A\x1A\x0A\x00\x00\x00\x0D\x49\x48\x44\x52\x00\x00'
-    '\x00\x10\x00\x00\x00\x10\x08\x06\x00\x00\x00\x1F\xF3\xFF\x61\x00\x00\x01'
-    '\x6C\x49\x44\x41\x54\x78\x9C\xA5\x92\x31\x8E\x02\x31\x0C\x45\x7F\xE2\x90'
-    '\x21\x1A\x8D\x44\x47\x47\x47\xC1\xA9\xB8\x05\x17\xE0\x04\x70\x03\x4E\x41'
-    '\x4B\x89\x38\x00\x07\x40\x42\x34\x14\x20\x21\x14\x0F\x93\x78\xAB\x64\x18'
-    '\x46\xCB\x16\x6B\xC9\x45\x9C\xF8\xE5\x7F\x27\x4A\x44\xF0\x9F\x30\x00\xB0'
-    '\x58\x2C\xE4\x7E\xBF\xC3\x39\x07\x63\x0C\x00\xE0\x13\x1C\x42\xC0\xF3\xF9'
-    '\xC4\x78\x3C\xC6\x72\xB9\x54\x79\x43\x44\xB0\xDD\x6E\x85\x99\xA5\x69\x9A'
-    '\xAF\xC9\xCC\xB2\xDB\xED\x64\xB5\x5A\x89\x88\x40\x44\xA0\xD7\xEB\xB5\x4C'
-    '\xA7\x53\xD4\x75\x0D\xEF\x3D\x98\x19\xCC\x0C\xEF\x7D\x2F\xEB\xBA\xC6\x64'
-    '\x32\x81\xF7\xBE\xB5\x70\xBB\xDD\x40\x44\x68\x9A\xE6\xAB\xD7\x64\x49\x29'
-    '\x05\xE7\x5C\x0B\x18\x8D\x46\x88\x31\x22\x84\x90\x0F\x7C\x36\x6A\xAD\x91'
-    '\x24\x03\x40\x8C\xB1\x05\x54\x55\xD5\x03\x28\xA5\x3A\x43\x4C\x0D\x31\x46'
-    '\x30\x33\xCA\xB2\x6C\x01\x5A\x6B\x78\xEF\x3B\x12\xDF\x1B\xB5\xD6\x88\x31'
-    '\xE6\x3A\x33\xA3\x28\x8A\x16\x50\x96\x25\x8A\xA2\xC0\x60\x30\xE8\x01\x94'
-    '\x52\x1D\xB9\x4A\x29\x84\x10\xF2\x53\x03\x80\x71\xCE\xC1\x39\x07\x22\x82'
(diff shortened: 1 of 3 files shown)


topic2k:
Hmm, i don't know if it is a good idea to completly remove system and hidden files. What if a plugin comes with and needs a special file, no matter if it is a system file or a (maybe accidently) hidden file.
Instead i would use a list of defined files to exclude.

kdschlosser:
if it's something that needs to be included with the plugin it shouldn't be a system or hidden file. and the person including said file should remove those attributes... 

NO file should be hidden that is shipped with a plugin..

topic2k:
Would like to have comments from others.... HELLO, anybody out there? 😄

kdschlosser:
I can see where you are coming from as well... 

but let me ask ya this...
why would a plugin developer want to keep a file hidden? or mark it as a system file? (which can also keep it hidden)

only thought that comes to my mind is they were trying to do something malicious with it.

ThomasBott:
Well, you can do malicious things with unhidden, not system files as well. If you ask me, less restrictions, less problems. I do not see any specific issues with filtering all system or hidden files at the moment though, but you never know what comes in the future or if it's useful for something.
But I do would very much welcome if the Thumb.db files get's filtered out, these files are so annoying!

topic2k:
What about this: Filter some known Files like thumbs.db etc. and if other hidden/System Files are found, inform the author that packaging those files might be rejected in a future version of EG.

Btw: what about doing that check even on unpacking/installing?

kdschlosser:
hey, I just came up with a base Idea. so whatever you all want to do we can do.. makes me no never mind. and a "system" file is marked as such for only one purpose. to be able to keep them hidden even if the show hidden file is turned on. 

well that is the only purpose for it i have ever seen. but a file marked as system functions just the same as a file not marked with system.

I have an idea based on the feedback. so let me do it up and see if it's a good.

ThomasBott:
Doing the check on unpacking / installing (in addition) would make a lot more sense, as you can otherwise just add system or hidden files afterwards with 7zip (or another tool).
It kinda depends on what we would like to achieve here, removing specific files like Thumb.db in the export / (import) is a convenience feature I'm sure everybody welcomes :)
Removing the other files could cause issues (not sure if it ever will though), but if we go for it, I like the idea with the notification

kdschlosser:
well this is my game plan. if hidden or system files are present. to have a dialog open that will list the files/directories with check boxes and have a list of known nuance files so those will automatically be unchecked and the rest checked check would mean to include the file. with a nice handy Ok or Cancel button. Cancel button would stop the whole process.

and i will also have some kind of a dialog to inform the person installing the plugin that there are hidden.system files that are included with said plugin.

that was my thought on it..

topic2k:
What is the status of this?

kdschlosser:
I have been trying to play catch up.. I will finish it up today.

kdschlosser:
I have to take a look at the code to make sure it is going to be compatible with the other PR for the plugin Install. I think it should be.. But i want to do it have a dialog appear if there are hidden or system files. and that dialog to have a check list in it and files that are checked will be added to the plugin package. so by default everything will be checked off except known nuisance items. like desktop.ini. and thumbs.db. i do not know if there are any other nuisance files you can think of..

I will also account for files/folders that start with a .

kdschlosser:
OOOOOOOOOOkkkk. give this whirly bird a twirl. 

I created that dialog that allows for the user to select the files/folders if there are any hidden, system or file that start with a '.'

eg/Classes/PluginInstall.py
```diff
@@ -283,3 +431,154 @@ def VisitName(self, node):
...
+                    )
+
+                    ff[key] = not bool(ff[key])
+                    self.tree.SetPyData(child, ff[key])
+                    self.tree.CheckItem(child, ff[key])
+
+        self.root = root = self.tree.AddRoot(self.fileFolders['sourcePath'])
+        self.tree.SetPyData(root, self.fileFolders)
+        ReadFileFolders(self.fileFolders, root)
+        self.tree.Expand(root)
```
topic2k: I would open all children:
`self.tree.ExpandAllChildren(root)`

kdschlosser: I can do that.

eg/Classes/PluginInstall.py
```diff
@@ -57,30 +61,174 @@
...
+
+def GetAttributeFlags(path, name):
+    path = os.path.join(path, name)
+    res = []
+    if name.startswith("."):
+        res.append('.')
+    if IsHidden(path):
+        res.append('HIDDEN')
+    if IsSystem(path):
+        res.append('SYSTEM')
```
kdschlosser: I can do that as well...

eg/Classes/PluginInstall.py
kdschlosser: And the thing with the gradient was something new. because I have not messed around with that specific wx widget before. I will change it to the stock look.

eg/Classes/PluginInstall.py
```diff
@@ -283,3 +431,154 @@ def VisitName(self, node):
...
+                wx.MAXIMIZE_BOX
+            )
+        )
+
+        headerBox = eg.HeaderBox(
+            self,
+            name='Plugin Install',
+            text=(
+                'There are hidden and system files please select\n'
+                'the files you wish to add to the egplugin package.'
```
kdschlosser: I have a question on this bit of it..

Do you think we should put a Text class into each bit we do. or should be put all core Text class items into eg.Text. because I have put everything into the core Text class so far. 

personally I would rather see a text class in each of the different files. for the various components. it makes it much easier when editing. because everything related to that specific piece is in one file. 

whatever you would like me to do.

topic2k: Well, to me it looks like eg.classes.Text is a relict before there was eg.classes.TranslateableStrings (not sure, haven't looked into svn history). I'll admit that you're right to put the texts into the corresponding files. So we should go that way (and get rid of eg.classes.Text).

kdschlosser: Fantastic

The other PR i did was for adding the more plugin install information.. I will add this to a new class inside of the file and then create a new PR to move all of the other plugin related text out of the Text file that is fi the other PR got merged. i think it did but i honestly do not remember

topic2k:
Found two bug:
- when unchecking a folder i'll get an AttributeError (NoneType has no IsOk()).
- the exported plugin file doesn't have the checked files/folders in the archiv.


For testing i have the following structure:
2017-01-27 17_33_54-plugin install - hidden_system files

PR #149, opened 2016-12-31
fix for the empty Logctrl after restoring from tray.

### LINKED ISSUE #138: Restoring EventGhost from tray results in an empty Log
In EG 0.5 beta5, Restoring EventGhost from tray results in an empty Log, the log can still be filled by new events and actions but the ones happened while EG was hidden (or before) are gone.

topic2k:
fixes #138


ThomasBott:
confirmed, thx!

PR #150, opened 2016-12-31
[System] - Bugfix - Allow multiple quoted arguments in calls to CMD via /S

edemaine:
This fixes the "Run Command" action of the System plugin to allow the command and one or more arguments to have quotes/spaces in them.  As reported at http://www.eventghost.net/forum/viewtopic.php?f=4&t=9491

Basically, `cmd /c` does not work well when there are multiple quoted arguments in the command line.  `cmd /s/c` always just removes the first and last quote, so by adding extra such quotes at either end, we prevent actual quote removal.  (As the code is currently, `cmd` sometimes removes the first and last quote, even if there are other quotes in between, which changes the pairing among quotes!)

### DIFF
--- a/plugins/System/Command.py
+++ b/plugins/System/Command.py
@@ -128,10 +128,10 @@ class Command(eg.ActionBase):
                 eg.folderPath.TemporaryFiles,
                 "EventGhost-output-%s.txt" % ttime()
             )
-            processInformation.lpParameters = '/C %s > %s' % (command, filename)
+            processInformation.lpParameters = '/S/C "%s" > %s' % (command, filename)
             processInformation.fMask = SEE_MASK_NOCLOSEPROCESS
         else:
-            processInformation.lpParameters = '/C %s' % command
+            processInformation.lpParameters = '/S/C "%s"' % command
         if runAsAdmin:
             processInformation.lpVerb = "runas"
         processInformation.nShow = 0
@@ -288,7 +288,7 @@ class Command(eg.ActionBase):
 
 def popen(cmd, si):
     return Popen(
-        'cmd /C %s' % cmd,
+        'cmd /S/C "%s"' % cmd,
         stdout = PIPE,
         stderr = open(devnull),
         startupinfo = si,


kdschlosser:
I am trying to replicate this problem

when i enter this 

    SOME DEIRECTORY WITH SPACES\nircmd.exe win child class "Shell_TrayWnd" show class "TrayClockWClass"

it works just fine. and there are multiple quoted arguments. but also the use of command i do not believe was for the intention of passing arguments. that would be the System\Start Application actions purpose as it does allow for setting arguments.

edemaine:
What should be happening in your example is that CMD will strip the first and last quote, leaving `SOME DEIRECTORY WITH SPACES\nircmd.exe" win child class "Shell_TrayWnd" show class "TrayClockWClass`
(I'm assuming you put "SOME DIRECTORY WITH SPACES\nircmd.exe" in quotes as well.)  It may be that it's working because the quoted arguments have no spaces in them, and it will fail to work if you do a command that has an argument with a space...  I'm not sure exactly, but I think you're getting lucky, and if you try `"C:\path with space\nircmd\nircmd.exe" setdefaultsounddevice "device with space"` it won't work without this PR.

My impression was that the goal of "Run Command" was to run any command-line command, as you would in CMD... That's at least it's current behavior, aside from this quoting issue.

You can read the effect of `/S` using `CMD /?`, included below.  Basically, the quote removal behavior is complicated, but `/S` makes it very simple: always remove the first and last quote.  By adding an extra first and last quote, as in this PR, it guarantees that CMD doesn't mess with quotes at all.

```
If /C or /K is specified, then the remainder of the command line after
the switch is processed as a command line, where the following logic is
used to process quote (") characters:

    1.  If all of the following conditions are met, then quote characters
        on the command line are preserved:

        - no /S switch
        - exactly two quote characters
        - no special characters between the two quote characters,
          where special is one of: &<>()@^|
        - there are one or more whitespace characters between the
          two quote characters
        - the string between the two quote characters is the name
          of an executable file.

    2.  Otherwise, old behavior is to see if the first character is
        a quote character and if so, strip the leading character and
        remove the last quote character on the command line, preserving
        any text after the last quote character.
```

topic2k:
> when i enter this
> 
>  >SOME DEIRECTORY WITH SPACES\nircmd.exe win child class "Shell_TrayWnd" show class   "TrayClockWClass"
>
> it works just fine. and there are multiple quoted arguments. but also the use of command i do not believe was for the intention of passing arguments. that would be the System\Start Application actions purpose as it does allow for setting arguments.

I would tend to say that Run Application is for GUI-programs and Run Command is for DOS/CLI-commands, because the latter lets you use the output of the command as payload, which isn't possible with the former. And that makes sense because i wouldn't expect an (CLI-)output from a GUI-App.

I was able to replicate the problem, so i think the PR should be merged.

Maybe we should think about merging the two actions in a future version. I did a quick look and it looks like the internal code is nearly the same except of the output capture for payload.

kdschlosser:
Personally for the time being I think a name change of the actions should be done until the 2 can be merged. I was playing around with it as well last night and i managed to replicate the problem as well. I didn't get a chance last night to test to make sure the PR worked properly. But if you tested that end of it then by all means it's a fix. but consider the changing of the names of the actions to something like "Run a Windows Program" and Run a Command Line Program".

I personally like how the arguments are separated from the command in the one version of it. But i think that having the 2 brought into one action is the best way to go.

PR #163, opened 2017-01-08
Migrate from `VistaVolEvents` to `AudioEndpointControl`

### LINKED ISSUE #109: Migrate from `VistaVolEvents` to `AudioEndpointControl`
Now that the wheels are in motion, I'm opening an issue to track and discuss this important addition. Thanks for taking it on, @ThomasBott and @jonisb!

topic2k:
Replace the current audio system with the AudioEndpointControl system from jonib and Sem;colon.
(I only did some code cleanup.)

closes #109

### DIFF
--- a/lib27/site-packages/AudioEndpointControl/EndpointvolumeAPI.py
+++ b/lib27/site-packages/AudioEndpointControl/EndpointvolumeAPI.py
@@ -1,0 +1,238 @@
+# -*- coding: utf-8 -*-
+
+from __future__ import print_function, unicode_literals, absolute_import
+from ctypes import (
+    POINTER as _POINTER,
+    HRESULT as _HRESULT,
+    c_float as _c_float
+)
+from ctypes.wintypes import (
+    BOOL as _BOOL,
+    DWORD as _DWORD,
+    UINT as _UINT,
+)
+from comtypes import (
+    GUID as _GUID,
+    IUnknown as _IUnknown,
+    COMMETHOD as _COMMETHOD,
+    STDMETHOD as _STDMETHOD,
+)
+
+from ctypes import Structure
+
+"""
+typedef struct AUDIO_VOLUME_NOTIFICATION_DATA
+    {
+    GUID guidEventContext;
+    BOOL bMuted;
+    float fMasterVolume;
+    UINT nChannels;
+    float afChannelVolumes[ 1 ];
+    }   AUDIO_VOLUME_NOTIFICATION_DATA;
+
+typedef struct AUDIO_VOLUME_NOTIFICATION_DATA *PAUDIO_VOLUME_NOTIFICATION_DATA;
+"""
+
+
+class AUDIO_VOLUME_NOTIFICATION_DATA(Structure):
+    pass
+
+AUDIO_VOLUME_NOTIFICATION_DATA._fields_ = [
+    ('guidEventContext', _GUID),
+    ('bMuted', _BOOL),
+    ('fMasterVolume', _c_float),
+    ('nChannels', _UINT),
+    ('afChannelVolumes', (_c_float * 1)),
+]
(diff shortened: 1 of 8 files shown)


ThomasBott:
> only did some code cleanup

Haha, you're funny, my diff is all colored! ^^

Some Remarks:
- in the System plugin, line 299: GetMute returns an integer while SetMute requires Bool. jonib wrote that GetMute should return Bool and will do that in the future, but currently it returns integer, so we should convert the output of that function to bool. The `GetMute() == 1` was there for a good reason.
- I also noticed that you don't import * any more in the AudioEndpointControl __init__.py. That's not a problem for how the library is used at the moment, but is an API change/limitation, as you can't use functions or constants any more that are not directly referenced in __init__.py (as you only import the ones that are referenced)

topic2k:
I tested it and always got 0 or 1 from GetMute. That is absolutly fine. Bool is just a special int. 0 is False, all other is True.

Regarding the * imports: are there any functions that need them maybe? If so, i will change it back. (i personaly don't like it much because it could lead to errors, that are not easely trackable. [shadowing/overwriting variables])

ThomasBott:
> Regarding the * imports: are there any functions that need them maybe? If so, i will change it back. (i personaly don't like it much because it could lead to errors, that are not easely trackable. [shadowing/overwriting variables])

Well, there are some used in the Example.py of the lib, wish is pretty much the documentation, maybe you can use it as a reference.

> I tested it and always got 0 or 1 from GetMute. That is absolutly fine. Bool is just a special int. 0 is False, all other is True.

Sure, if you let python interpret the integer as bool you always get False on 0 and True on 1, but if you like to use the result of the GetMute function as parameter for the SetMute function, you will get a type error! Also it's confusing for some other scenarios I can think of.
Also, the returned type in the lib will change to bool in the future, to preserve backwards compatibility with as it is now, we will have to do an ugly bool to Integer conversion as otherwise the function would return a different data type than before and this can brake backwards compatibility (it will with one script I have).

jonisb:
I'm checking the library code and have done some changes as there was some code for testing and commented code that don't need to be there.
I'll compare the changes done here when I'm done.

I'll look into the GetMute bool issue.

> Regarding the * imports: are there any functions that need them maybe? If so, i will change it back. (i personaly don't like it much because it could lead to errors, that are not easely trackable. [shadowing/overwriting variables])

I'll check the changes to the importing and maybe change my code.

jonisb:
Remove "**MMDeviceAPILib.py**" and "**stdole.py**" as those are only needed for Python 2.6, in later version they are automatically generated. (Python 2.6 can't generate the code properly)

The ImportError exception in the code is to generate "**MMDeviceAPILib.py**" when not provided locally. I have it there so my library can support EventGhost 0.4.

I kinda want my library to be included and not an edited version as this library is NOT only going to be used in EventGhost, so I would like if these changes ware made to the [AudioEndpointControl](https://github.com/jonisb/AudioEndpointControl) repository first and then committed to the EventGhost repository. I would appreciate a pull request.

I might not be up to checking all the code today as there was more changes then I anticipated.

topic2k:
Ok, didn't know about the github repo. If you make a current release, i will add it as dependency instead of adding it to the EG source.

jonisb:
OK, I thought you knew, but I guess that fact isn't as clear as I thought from the Library announcement in the EventGhost forum heh.

I'll try to make the changes soon.

kdschlosser:
I wanted to make note in here for this so that way it is documented.

There needs to be better handling of a device that is not attached to the system anymore. If an action gets performed on a device that is no longer available it causes EG to lockup and crash. Use case is with remote desktop if you login using RDP the device is no longer available a virtual device has taken it's place. I know the session can be tested to see if it's a remote session. what i do not know is if EG is running in another session can make be called to make the changes or does windows freeze all running programs when it logs the user off.

PR #165, opened 2017-01-08
[EventGhost] eg.Environment, Adds parameter "variable" to method Get

kdschlosser:
Adds parameter "variable" to the Get method in eg.Environment. it completes the convenience method. It is defaulted to None, so if no variable is specified it will do what it was originally programmed to do and return a sorted list of the variables. If a variable name is specified it will return that variable if it has been set otherwise it will return None.

### DIFF
--- a/eg/Classes/Environment.py
+++ b/eg/Classes/Environment.py
@@ -58,8 +58,11 @@ class Environment:
                 del os.environ[var]
 
     @staticmethod
-    def Get():
-        return Environment.Sort(os.environ)
+    def Get(variable=None):
+        if variable is None:
+            return Environment.Sort(os.environ)
+        else:
+            return os.environ.get(variable, None)
 
     @staticmethod
     def GetLatest():


kdschlosser:
may want to hold on of you are going to merge this...

I was looking at eg.Environment and I think it could be improved upon and seeing how there is nothing in the core that uses it except to refresh it. which I would leave the same.

I am thinking instead of AppendPath() or Get() or Set()

it could be as easy as.

eg.Environment.User.PATH += 'C:\\SOME_PATH'
eg.Environment.System.PATH += 'C:\\SOME_PATH'

for any path type variables __setattr__ would bring up the path. compare the 2 to see what is being added or removed and set the ';' properly. and for any other variables if it is not in the protected list of environment variables it would either create a new one or change an existing. 

the 2 types of environment variables really need to be separated. the reason why if because there is a blacklist of specific variables not to be set in the Users.

I also have a complete list of all the known hidden or otherwise windows default variables. and which ones can be changed and which ones can't. another nice thing

but I also feel that a snapshot of all the existing variables should be taken when EG starts and when EG closes anything added needs to be removed and all the originals need to be put back the way they were.

and some of the methods would stay like Refresh() but there would be 3 of them
one in System
one in Users
and one in Environment that would initiate the other 2

let me know what ya think...

topic2k:
> let me know what ya think...

As long as it don't break anything else, go for it :)
The only concern i have, is with this:
> but I also feel that a snapshot of all the existing variables should be taken when EG starts and when EG closes anything added needs to be removed and all the originals need to be put back the way they were.

You should also take care if the variables were modified not by EG between start/close of EG. And maybe add an option to restore the variables, just in case the user wanted the changes to persist.

kdschlosser:
EG will keep a running knowledge of what it has changed. and when reverting if the current setting is in fact not the same it will not touch it..

kdschlosser:
or maybe add an option to the options dialog. that will be a simple check box.  titled
 "Leave changes made to environment variables after close."

But even if we don't do that bit of it.. it still really really really needs to be separated into User and System. With a set of blacklisted variables for each..

PR #171, opened 2017-01-11
add six to included modules

### LINKED ISSUE #169: System Plugin Error
OK so this is the skinny. 
Youi have to do this on a computer that does not have Python installed on it. it's an intertwining issue

But if you grab any of the recent appveyor builds and install it you will be met with some ugly red.
but this error is very misleading. and it took me a while to track down the problem...

but this is how i managed to find the source of it.

if you open the system plugin __init__ and move all of the import to below the register plugin. and leave the import eg above it

then run eventghost you will get the true source of the problem. 

    Error loading plugin file: C:\Program Files (x86)\EventGhost\plugins\System
    Traceback (most recent call last) (WIP-2017.01.03-18.17.05):
      File "C:\Program Files (x86)\EventGhost\plugins\System\__init__.py", line 57, in <module>
        from qrcode import QRCode, constants as QRconstants
      File "C:\Program Files (x86)\EventGhost\lib27\site-packages\qrcode\__init__.py", line 1, in <module>
        from qrcode.main import QRCode
      File "C:\Program Files (x86)\EventGhost\lib27\site-packages\qrcode\main.py", line 1, in <module>
        from qrcode import constants, exceptions, util
      File "C:\Program Files (x86)\EventGhost\lib27\site-packages\qrcode\util.py", line 4, in <module>
        import six
    ImportError: No module named six

 I know there was someone in the forum complaining of this issue. and again a second person mentioned the issue but had said that they tried with appveyor builds as well. so i started snooping and installed an appveyor build. and low and behold there was the problem...

now I do not believe we are not experiencing these problems because we have six installed into python and not into EG but EG will still load it because EG now grabs the python site packages

which is why i think it is a horrible idea to have it set up to do this. and I do believe I have mentioned this kind of an issue happening because of it. Package Installer built into EG is the way to go.

I have not gone through the legwork of making a fix for it as of yet. but I am sure it is as simple as adding six to the dependencies.

topic2k:
fixes #169

### DIFF
--- a/_build/Build.py
+++ b/_build/Build.py
@@ -48,6 +48,7 @@ class MyBuilder(builder.Builder):
         "pkg_resources",
         "pythoncom",
         "pywin32",
+        "six",
         "win32com",
         "wx",
     ]


kdschlosser:
yeah adding it to the included modules will work as well.

but i do not know how to do a review. i need instruction. TY

topic2k:
Well, basically it means take a look at the code, test it and say what you think about it. 😄
But here is short [How To Do](https://help.github.com/articles/approving-a-pull-request-with-required-reviews/)

PR #172, opened 2017-01-11
[Speech] - Bugfix - Fixes pythoncom.com_error when using pywin32 220 and also corrects Text usage

kdschlosser:
This fixes the way Text was being used. this way it will not create an infinite loop

Fixes win32com/pythoncom.com_error when using pywin32 220

### DIFF
--- a/plugins/Speech/__init__.py
+++ b/plugins/Speech/__init__.py
@@ -81,21 +81,6 @@ import pythoncom # NOQA
 from time import strftime # NOQA
 
 
-def _GetDescInvokeType(entry, invoke_type):
-    if not entry or not entry.desc:
-        return invoke_type
-    varkind = entry.desc[4]
-    if (
-        varkind == pythoncom.VAR_DISPATCH and
-        invoke_type == pythoncom.INVOKE_PROPERTYGET
-    ):
-        return pythoncom.INVOKE_FUNC | invoke_type
-    else:
-        return varkind
-
-win32com.client.dynamic._GetDescInvokeType = _GetDescInvokeType
-
-
 class Text:
     suffix = "SpeakingFinished"
     ttsError = 'Speech: Unable to start the TTS engine'
@@ -454,3 +439,4 @@ class TextToSpeech(eg.ActionClass):
                 volumeCtrl.GetValue(),
                 devChoice.GetStringSelection()
             )
+
--- a/plugins/Speech/__init__.py
+++ b/plugins/Speech/__init__.py
@@ -22,6 +22,9 @@
 #
 # Changelog (in reverse chronological order):
 # -------------------------------------------
+# 2.6 by K         2017-01-11 13:23 UTC-7
+#     - bugfix - text class not being utilized properly
+#     - bugfix - win32com com_error
 # 2.5 by blackwind 2016-05-19 18:40 UTC-7
 #     - bugfix - instantiate Text class in declarations
 # 2.4 by Pako 2016-03-05 09:44 UTC+1
@@ -39,15 +42,15 @@
 #     - initial version
 
 eg.RegisterPlugin(
-    name = "Speech",
-    author = "MonsterMagnet",
-    guid = "{76A1638D-1D7D-4582-A726-A17B1A6FC723}",
-    version = "2.2",
-    description = (
+    name="Speech",
+    author="MonsterMagnet",
+    guid="{76A1638D-1D7D-4582-A726-A17B1A6FC723}",
+    version="2.6",
+    description=(
         "Uses the Text-To-Speech service of the Microsoft Speech API (SAPI)."
     ),
-    url = "http://www.eventghost.org/forum/viewtopic.php?f=9&t=6828",
-    icon = (
+    url="http://www.eventghost.org/forum/viewtopic.php?f=9&t=6828",
+    icon=(
         "iVBORw0KGgoAAAANSUhEUgAAABAAAAAQCAYAAAAf8/9hAAAACXBIWXMAAA7DAAAOwwHH"
         "b6hkAAADh0lEQVR4nIXSW0yTBxgG4P/GxRkZ2+gWtR2YCBoOlsL/87fQg4WldkLHqQZo"
         "RcbBUBDCYFxsukaJqyLdAgw3p1AcOhHXNhzaIWLXtVSoWNx0LMsssspGUMYWMB3URVre"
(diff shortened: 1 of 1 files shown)


kdschlosser:
the audio end of things needs to be tested as i only have one audio device in this computer, would be great if someone could do that for me

plugins/Speech/__init__.py
```diff
@@ -383,7 +427,7 @@ def OnButton(event):
             (sizer2, 0, wx.EXPAND|wx.ALL, 5),
         )
         ACV = wx.ALIGN_CENTER_VERTICAL
-        sizer3 = wx.FlexGridSizer(0, 2, 5, 5)
+        sizer3 = wx.FlexGridSizer(0, 0, 5, 5)
```
topic2k: Leave the column count at 2 else all widget will be in one row.

eg-speech

per1234:
These unrelated changes should be broken into separate commits, if not separate pull requests:
- Correct Text usage in Speech plugin
- Make Speech plugin compatible with pywin32 2.20
- Speech plugin code formatting

Also, shouldn't the commit adding pywin32 2.20 compatibility also update https://github.com/EventGhost/EventGhost/blob/d8efe89d0d71b1020c8d63db8622a036463238d6/.appveyor.yml#L26?

plugins/Speech/__init__.py
kdschlosser: ok this newest commit should be good to go. I found the error in win32com and did a monkey patch to correct it. I did pass along the information to Mark Hammond. but form looking at the support ticket history i do not think there will be a fix for a longgggg time

kdschlosser:
and that is also one other thing.. the pip install of pywin32 doesn't work properly.. it is posted in one of the support tickets for pywin32

it was almost a complete rewrite. not something to break into different commits
everything got changed in it..

per1234:
>it was almost a complete rewrite. not something to break into different commits

I disagree. It could easily be broken into the three parts I listed above. The Text issue is unrelated to the pywin32 issue. The code formatting is purely aesthetics, not required for any sort of functionality. The only thing these changes have in common is they are all done on the Speech plugin.

Don't get me wrong, I very glad if the Text issue gets fixed, whichever way that happens. I'm also glad if the issue with pywin32 is fixed, having to downgrade didn't strike me as the best solution, though it did fix the bug. PEP8 compliance is also an excellent change. But I think it's very important to do atomic commits as [specified in CONTRIBUTING.md](https://github.com/EventGhost/EventGhost/blob/master/CONTRIBUTING.md#commits-count). This will make troubleshooting much easier, make it clear what was the purpose of each change, and make it easier for us to review the changes you are proposing. It also makes it less likely for a common sense, obviously beneficial change to be delayed or rejected due to being mashed together with a more controversial change.

kdschlosser:
The purpose to a commit would be for submitting "completed" code. doing one piece that will not function properly in it's self and committing it. then have to go back to change other thing and make a commit for that and still not have something functional  then having to go back to change how it's formatted, to go  back again to do something else. that's kind of the definition of stupidity.

and it makes for a mess. get a running rough draft. then commit. then fix whatever mistakes there are or bugs that are found and make commits for those. 

i start at the top. and work my way down. whatever is not right and needs to be changed gets changed. i don't keep on going back over it again and again. that's how mistakes get made and things get missed or forgotten about.

and I am not going to submit changed to something like the Text class that has the probability of not functioning properly until i finish the "whole" rewrite 

now what happens if i go through and reformat.. then commit... but oops i missed some things. then go fix those and commit again... crap still missed some things another commit...
that's bonkers. and if you revert your commit that just makes even more confusion.

if your intention is specifically to clean the code. then yes by all means. but having a commit for for each and every single change is a little much. that is why there is a description field to a commit.. to explain what has been done.

and as far as appveyor,  and also the dependency.  that does not have to be changed. it will still function the same. tho it can be changed. this is to have the powers that be go over it to make sure it runs right. then do any final last steps and complete any changes. there are things in which i am not able to test. as this pull request shows. and there were problems. that lead to other problems that lead to a lot of code changes. and if i had specifically done them like you state i would be going over work i just did only for the sole purpose of a commit which would end up being meaningless anyways because of the code changes.

kdschlosser:
welll.... as it turns out.... there is a plugin that was made by pako that uses the dde client from pywin32.

and this has not worked since version 214. and was reported as a bug in 2010 and has not been fixed at all.  I have corrected this plugin i am just waiting back on a response to see if it is working properly. but since 7 years has past with no resolution to it. i fear that unless we patch things like this our selves we will never be able to update the thing. it looks like version 214 was the version with the smallest number of problems. I am not sure if there is a specific reason as to why we have upgraded. I am sure that because of windows 10 there are some new elements. Oh and there is one other... in the system plugin pywin32 does not turn on a monitor properly in windows 8 and up. there is another that is pywin32. this has not functioned properly even in EG .4 releases. is there a possible alternative to pywin32?? or should i create a working dde and add it to EG as a module?  don't know what the resolution would be here.. there are ways of fixing things with pywin32 that work. but like the dde issues this has to deal with manifest files that are compiled into specific files. and I am unable to monkey patch that.

plugins/Speech/__init__.py
```diff
@@ -75,10 +75,25 @@
...
-from time import strftime
-import threading
-import pythoncom
+import win32com # NOQA
+import threading # NOQA
+import pythoncom # NOQA
+from time import strftime # NOQA
+
+
+def _GetDescInvokeType(entry, invoke_type):
```
topic2k: Maybe we should put those function in a separate file, something like eg.patches or eg.compatibility, in the case there will be more of those bugs? (dde?)

kdschlosser:
I was thinking the same thing as well...

the dde one is a little more complicated... in that i only reproduced the commands that the xmplay plugin actually used.. I don't know how much that specific modules is used. but I guess we can cross that bridge when we get there.

so what i will do is create a new EG file specific to the task of monkey patching pywin32

kdschlosser:
I have set the patching so it is in it's own folder under WinApi and each patch should be made into it's own file... and the name of the file is the name of the module being patched and inside the file would be the function or class being patched. or it's the whole module being replace as with the case of dde

and in the __init__ file the only thing that goes in there are the imports of the patches and setting the patch into sys.modules

that is about as easy as I could make it..

topic2k:
Hej K, will you redo the commits for a clearer commitlog? I [prepared ](https://github.com/EventGhost/EventGhost/compare/master...topic2k:(Speech)_reworked?expand=1)it so you nearly need just copy paste. I think you should make it so that your name shows up in the  EG changelog for this PR.

kdschlosser:
dude i will make a mess of it... go ahead and do it.. i don't care about the credit on this... it was really a group effort to be honest with ya... don't know if you can manually change the changelog or not.

topic2k:
ok, then i will make a new PR, it's the easiest.

kdschlosser:
OK. however you want to do it.

PR #177, opened 2017-01-15
Add module for patching pywin32

kdschlosser:
Adds a separate module for handling pywin32 monkey patches.

Added dde patch
Added win32com.client.dynamic._GetDescInvokeType

### DIFF
--- a/eg/Init.py
+++ b/eg/Init.py
@@ -76,6 +76,8 @@ def ImportAll():
     Traverse(eg.corePluginDir, "eg.CorePluginModule")
 
 def Init():
+    import WinApi.pywin32_patches # NOQA
+
     if eg.startupArguments.isMain or eg.startupArguments.install:
         import WinApi.COMServer  # NOQA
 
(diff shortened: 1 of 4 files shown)


topic2k:
Now with the patches should we use pywin32 version 220? (or 219 or 214 or ... ❔ 😕 ) (I'll wait for an answer before merging)

kdschlosser:
214 is old from 2010. anything windows 8+ related will not be in it...

and being that windows 10 basically just got released. 220 is bound to have more support for that. 

it's a coin toss... I haven't come across any other issues. and dde is a dead way of communicating with applications in favor of COM.  so i can understand as to why it hasn't been fixed..

and no one has posted a bug about the other issue as of yet.. so i posted one as well as the fix for it.. which may make them fix the problem quicker.

i say stick with 220 only because we haven't had any "major" issues with it. and we have been able to fix the ones we have come across. Big reason why is windows 8 and 10. there are things that changed as you can see from the power notifier i had to do some version checking there to get the right guid.

PR #182, opened 2017-01-20
[EventGhost] - Enhancement AddEvents, Adds events to info.eventList

kdschlosser:
changes the AddEvents to add events to the list instead of replacing it. but it also checks to see if the information in the list that has been passed is event, description and if it is only an event it adds a blank description.

### DIFF
--- a/eg/Classes/PluginBase.py
+++ b/eg/Classes/PluginBase.py
@@ -148,7 +148,15 @@ class PluginBase(object):
         self.info.actionGroup.AddActionsFromList(theList, defaultAction)
 
     def AddEvents(self, *eventList):
-        self.info.eventList = eventList
+        if self.info.eventList is None:
+            self.info.eventList = ()
+
+        for event in eventList:
+            if isinstance(event, basestring):
+                event = (event, '')
+            elif isinstance(event, list):
+                event = tuple(event)
+            self.info.eventList += (event,)
 
     def AddGroup(self, name=None, description=None, iconFile=None):
         """


eg/Classes/PluginBase.py
```diff
@@ -148,7 +148,15 @@ def AddActionsFromList(self, theList, defaultAction=None):
...
-        self.info.eventList = eventList
+        if self.info.eventList is None:
+            self.info.eventList = ()
+
+        for event in eventList:
+            if isinstance(event, basestring):
+                event = (event, '')
+            elif isinstance(event, list):
+                event = tuple(event)
+            self.info.eventList += (event,)
```
topic2k: Instead of redefining `event`, directly add to `eventList`?
```python
        for event in eventList:
            if isinstance(event, basestring):
                self.info.eventList += ((event, ''),)
            elif isinstance(event, list):
                self.info.eventList += (event, )
```

kdschlosser: yes. the System plugin is one example.. if you wanted to make more then one call to AddEvents because you wanted to keep events in different files for the purpose of code organization. you couldn't before... because it would just overwrite it.. I also added a little bit of lazy to it as well.. so if the plugin dev just wants to pass a list of events because they do not want to add descriptions this will reformat it to work properly. idk it was simple to do.

but here is another example... as events occur. if it is not in the info.eventList it can be added dynamically with a simple method call.  instead of having to code this out. Like when you scan the tree in the AddEventsDialog.. that way you have a mechanism to add them to the list easily. but also as i mentioned above.

topic2k: ehm, wrong conversation? Missunderstanding? I just asked for a little codechange. ☺️

kdschlosser: so if you come across an event and want to add it... all you would have to do is pass the eventstring to AddEventList..

now i should have it check for possible duplicates. that would also be beneficial. because if you use this method from multiple spots you won't have to add repeat code. just pass the event and let the method figure out if it should add it or not...

a dictionary might be a better container instead of a tuple for this... but i do not know how much code you would have to change with the AddEvents dialog to account for this. 

maybe the EventInfo class is what should really be used to house the data..  and can be iterated over to spit out keys and values of a dict that holds all of the events. set it up to do this out of the gate.. then you won't have to add them to add each event to an EventInfo class each time the AddEvent dialog loads.

kdschlosser:
it's funny because they are tied together...

this mechanism does not only apply to the person making the plugin..
you can call it as well when you are making the AddEvent dialog...
have it do what it needs to do all using common code. 

so as an example... you use the EventInfo class in the add event dialog.. 
make this thing so that as items are added it creates the EventInfo instance and stores that instead.
so instead of having to scan the whole tree.... have the EventItem class create an EventInfo class (if it does not already exist) and add it to the eventList.. then all you have to do is load the list with the AddEventDialog. no need to scan anything additional.. because things were added at time of creation.

yes no?? 

this is the central spot that the EventItem, AddEventsDialog, EventGhostEvent and any things else can access to add items to the list that may not be there. and do any instance checkings and duplicate checking here. so code doesn't have to be duplicated.

it might be easiest to just code up a quick example of what my melon is thinking.. I think it would make the add events dialog code smaller. but it would also allow for adding of items. whether is be a "static" entry or a "dynamic" entry for anywhere inside of EG

eg/Classes/PluginBase.py
kdschlosser: t's funny because they are tied together...

this mechanism does not only apply to the person making the plugin..
you can call it as well when you are making the AddEvent dialog...
have it do what it needs to do all using common code.

so as an example... you use the EventInfo class in the add event dialog..
make this thing so that as items are added it creates the EventInfo instance and stores that instead.
so instead of having to scan the whole tree.... have the EventItem class create an EventInfo class (if it does not already exist) and add it to the eventList.. then all you have to do is load the list with the AddEventDialog. no need to scan anything additional.. because things were added at time of creation.

yes no??

this is the central spot that the EventItem, AddEventsDialog, EventGhostEvent and any things else can access to add items to the list that may not be there. and do any instance checkings and duplicate checking here. so code doesn't have to be duplicated.

it might be easiest to just code up a quick example of what my melon is thinking.. I think it would make the add events dialog code smaller. but it would also allow for adding of items. whether is be a "static" entry or a "dynamic" entry for anywhere inside of EG

kdschlosser: i meant to say have the EventItem just pass it's name to the AddEventList and the AddEventList will check to see if it exists and if not it will add it

PR #184, opened 2017-01-23
Dialog gets stuck if a traceback occurs during it's creation.

kdschlosser:
if a traceback occurs during the creation of a dialog. a notification was already sent out to add the dialog to mainframes openDialogs. the FinishSetup method in TaskletDialog never gets to run there for never setting the setupFinished flag. so when the exception gets caught and printed and should happen along it's merry way there is some checking of the setupFinished and the destruction on the dialog only occurs when setupFinished is set to True. so a notification never gets sent to remove the dialog from the list. causing the "phantom dialog" effect. forcing you to kill the EventGhost process because the main frame thinks there is still an open dialog. 

Upon fixing this issue I happened upon 2 others caused by thread timing. The 2 things that were happening in 2 different threads are the selecting of the treeItem and the deleting of the treeItem. so what would happen is in the middle of the treeItem being selected it would get deleted. causing additional tracebacks.. which should not have happened. those are fixed as well..

### DIFF
--- a/eg/Classes/TaskletDialog.py
+++ b/eg/Classes/TaskletDialog.py
@@ -162,7 +162,6 @@ class TaskletDialog(wx.Dialog, eg.ControlProviderMixin):
             self.Configure(*args, **kwargs)
         except:
             eg.PrintTraceback()
-
         self.__done = True
         self.__resultsChannel.send((None, None))
         self.Destroy()
--- a/eg/Classes/TaskletDialog.py
+++ b/eg/Classes/TaskletDialog.py
@@ -162,10 +162,10 @@ class TaskletDialog(wx.Dialog, eg.ControlProviderMixin):
             self.Configure(*args, **kwargs)
         except:
             eg.PrintTraceback()
+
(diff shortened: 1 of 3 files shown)


eg/Classes/TaskletDialog.py
```diff
@@ -162,10 +162,10 @@ def ProcessingTask(self, *args, **kwargs):
             self.Configure(*args, **kwargs)
         except:
             eg.PrintTraceback()
+
```
ThomasBott: Did you add this line on purpose?

kdschlosser: did i add what on purpose?

kdschlosser: if you mean the blank line.... if there was a "blank line" there before and it had been tabbed out.. my IDE will remove any unneeded white space so it will look like an addition..

If memory serves in this case i was modifying the code right in that area so it is possible i added something then deleted it and didn't remove the empty line.. Let me go back and look at the code..

PR #199, opened 2017-02-19
Fixes log being scrolled to the top after restored from tray icon

### LINKED ISSUE #198: Event log is scrolled to the top after restoring EventGhost from tray
I have found a small issue: After EventGhost is restored from tray, the event log is always scrolled to the very top, which means you need to scroll manually to the bottom every time, to see the latest events.
EG 0.5 beta 6

kdschlosser:
Fixes #198 
I also added a special feature to this. it will record the last position of the log before minimized to the tray (destroyed) and it will return to that position in the log when it is restored (recreated)

### DIFF
--- a/eg/Classes/Document.py
+++ b/eg/Classes/Document.py
@@ -65,6 +65,7 @@ class Document(object):
         self.frame = None
         self.reentrantLock = Lock()
         self.expandedNodes = set()
+        self.visibleLogItem = 0
 
     def AfterLoad(self):
         if (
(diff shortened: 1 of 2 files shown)


topic2k:
One wish: If autoscroll is enabled, don't scroll to last visible item, because there could be meantime new log entries: [Here i made the nesseccary changes.](https://github.com/kdschlosser/EventGhost/compare/bugfix-scrollaftermaximize...topic2k:bugfix-scrollaftermaximize)
If you agree with me to handle it that way, please add it to the PR.

kdschlosser:
I get what you are saying.. you want it to scroll all the way to the bottom if that is where it was even if events have taken place while it was minimized. that can easily be done at time of recording the actual item number to scroll to and bounce it against GetItemCount()

PR #200, opened 2017-02-22
[System] - Enhancement - Adds device names to events. Adds additional payload data

kdschlosser:
This is a first draft and I will be adding to it. I wanted to create a PR for it so it can get tested and if there are any suggestions they can be made.

Adds device names to the System.DeviceAttached, System.DeviceRemoved, System.DriveMounted and System.DriveRemoved events.

Added a payload with metrics about the device. the metrics vary per  device type


kdschlosser:
I still have yet to add the docs on this. but the get device action seems to work well. I am unable to specify searchable attributes for all the various kinds of devices as i do not have all the different types available. but for a good chunk of them they are good to go and i used an attribute that all the various kinds have even tho the information inside of it might not be the best to use. but this can be added to as time goes on.

kdschlosser:
OK I am going to need some help with this. I seem to have 2 issues so far. and I am unable to figure out why it is doing it.

One is with the unmounting of a network drive.. it throws no event.. Now it does trigger the notification.. and i store the instance of the mapped drive as well as the drive letter.. and they are there but for some reason they disappear.

the second is the help files for the dialog.. there seems to be some kind of a memory issue when loading the attribute.

the help files are compressed into a zip file (due to the raw size of them i didn't want to add a whole meg to the size of EG for the sake of help files that will only get loaded when the GetEvent dialog gets opened. So i add the location of the help.zip to sys.path.. this way i can import it.. the help is one file with attributes set in it for each of the help topics.. and i do a getattr to grab the proper attribute.. what seems to happen is the proper attribute gets loaded and displayed.. but at the end a bunch of garbage displays and then after that a completely different help attribute is displayed. I am not sure as to what is causing this.. I have been forced to use the old version of iewin included with WX this is due to a dynamically created activeX control that is made when you install wx this does not get done with EG and is in the excluded modules list. and I have also tried the html control that is in EG and this does not work with the html code that is in the help. I am not sure if there is a way to fix this issue..

if someone could lend a helping hand that would be great

kdschlosser:
I have fixed the Help. and sorted some other minor bugs... still have to finish up the docstrings. and have the mapped drive removed bug to tackle. But this is all coming together. any suggestions are welcomed so please voice your opinion.

kdschlosser:
OK the network drive issue is resolved.. and I have finished adding the last of the doc strings...

so unless someone can find something wrong with this or any bugs.. or wants to suggest something to add.. speak now...

Otherwise this has got the DONE stamp on it.

kdschlosser:
OK i hammered out a few more glitches.. But I have also sped up the drive mounting. Instead if iterating through all of the drives i use the WMI SQL query to select the drive based on the drive letter.

I also fixed an issue where the wrong drive type was being selected. but i also added Removable Drives. and Ram Drives to the list of drive types.

plugins/System/DeviceChangeNotifier.py
```diff
@@ -33,105 +47,1396 @@
...
+    ' WHERE AssocClass=Win32_%s'
+)
+
+
+def _parse_vendor_id(vendor_id):
+    """
+    Parses the vendor id that is received from a Windows notification.
+
+    :param vendor_id: str() Notification vendor id
+    :return: str() Modified vendor id.
```
kdschlosser: yes sir  I can do that...

plugins/System/DeviceChangeNotifier.py
```diff
@@ -1236,16 +1229,12 @@ def __call__(self, pattern=None, **kwargs):
...
-
-        :param kwargs: Keyword must be a valid device type. Value is the string
-         to search for.
-
         GetDevices(DiskDrive='SOME HDD NAME')
 
-        :return: tuple() WMI instances that represent a device.
+        :type pattern: str
+        :type kwargs: dict
+        :rtype: tuple
```
topic2k: Hmm, seems like we misunderstood us here. I didn't meant to remove the description of the parameter, just change the way the type is declared. I'm sorry for my vague description.

Instead of writing `str()` before the description:
```rest
:param pattern: str() text to search for.
:param kwargs: Keyword must be a valid device type. Value is the string to search for.
:return: tuple() WMI instances that represent a device.
```
add a line with the type for the parameter:
```rest
:param pattern: text to search for.
:type pattern: str
:param kwargs: Keyword must be a valid device type. Value is the string to search for.
:type kwargs: str
:return: WMI instances that represent a device.
:rtype: tuple of WMI
```
or alternative put the type before the parameter name:
```rest
:param str pattern: text to search for.
:param str kwargs: Keyword must be a valid device type. Value is the string to search for.
:return: WMI instances that represent a device.
:rtype: tuple of WMI
```
---
**To show the difference** (i've removed the long description here to make the pics smaller):

your first version:
grafik

with my changes:
grafik

what you have now:
grafik

kdschlosser:
ok I will make the changes... But this is going to take a little while. I am going to be changing how I am accessing WMI. I am going to be ditching the majority of a looping code I have learned how to use WMI Query Language (WQL) and simply ask it for what I want. I feel this will improve performance.. I also know that I can set up specific device notifiers. I am going to do a bit more reading on that end of things. I could attach to the device and have it notify me if it gets removed. or if something changes with it, like ejecting a disc from a ROM drive. that will really simplify the code. I also learned of some additional parameters that can be passed when creating the WMI instance. to change how they are handled, I have to do some more reading on that end of things. I also want to change out the payloads to just pass an instance of the device.. and the user can then just ask that instance for the different things they want. I think it will keep the log cleaner looking.  

I do not know if you have checked out that other pull request for the network adapter enabling and disabling. but I am also going to make one for BT devices.

kdschlosser:
I made some more changes.. I have improved the speed of the event generation by a vary large amount. I also had found a guid for network cards that is not documented so I added that as well. and also squished some duplicate events being fired

kdschlosser:
doc strings should be fixed now.. let me know if there is anything else that needs to be done.

kdschlosser:
Going to remake. do not want to use pywin32

PR #204, opened 2017-02-26
[System] Adds Windows XP support and moves the registering of the GUID's to the main thread

kdschlosser:
Windows XP does not support the use of RegisterPowerNotification and UnregisterPowerNotificatrion it always sends the notification. I want to thank Diz for providing the solution to fix this. Tho I did modify it so that in the future any new Windows versions will force a review of the code. I did this because of how Microsoft likes to change things about. We do want to make sure we have a working plugin. So in the event that there is an unsupported GUID in a new OS it will not register the GUID's because all of the OS checks will return False. 

The Registering of the GUID's is much happier being done in the main thread. even tho it does appear to work if done in another thread, it does not function properly. I discovered this issue when working with another part of EG that also uses win32com and when i had an error while working with that it would also cause the PowerBroadcastNotifier to throw a traceback. and only the PowerBroadcastNotifier would even tho the use of win32com is use throughout EG. and one of the traceback has some information in it about not being run in the main thread. 

And now since it is done in the main thread there are a couple of new events on startup as well. this is a good indicator it is functioning properly. And I believe this is the fix for #181. Not setting those GUID's in the main thread really mucks up odd parts of pywin32. I feel this is something that we may need to monkey patch pywin32 to have it throw an exception if it is not being done in the main thread as to avoid any future issues, IE: my forgetfulness or even a plugin dev. I suggest this because there is no indication there is a problem just random weird things happening.

### DIFF
--- a/plugins/System/PowerBroadcastNotifier.py
+++ b/plugins/System/PowerBroadcastNotifier.py
@@ -62,29 +62,8 @@ SVR_ON = 0x1
 AWY_EXITING = 0x0
 AWY_ENTERING = 0x1
 
-PI = wx.PlatformInformation()
-# Windows 10 /  Server 2016
-WIN_10 = PI.CheckOSVersion(10, 0)
-WIN_8 = False
-WIN_7 = False
-WIN_VISTA = False
-WIN_XP = False
 
-if not WIN_10:
-    # Windows 8, 8.1 / Server 2012, 2012 R2 / RT, RT 8.1
-    WIN_8 = PI.CheckOSVersion(6, 2) or PI.CheckOSVersion(6, 3)
-if True not in (WIN_10, WIN_8):
-    # Windows 7 / Server 2008 R2
-    WIN_7 = PI.CheckOSVersion(6, 1)
-if True not in (WIN_10, WIN_8, WIN_7):
-    # Windows Vista / Server 2008
-    WIN_VISTA = PI.CheckOSVersion(6, 0)
-if True not in (WIN_10, WIN_8, WIN_7, WIN_VISTA):
-    # Windows XP x86, x64 / Server 2003, 2003 R2 / Tablet PC, MCE
-    WIN_XP = PI.CheckOSVersion(5, 1) or PI.CheckOSVersion(5, 2)
-
-
-if True in (WIN_8, WIN_10):
+if eg.WindowsVersion >= 8:
     GUID_CONSOLE_DISPLAY_STATE = GUID(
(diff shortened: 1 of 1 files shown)


kdschlosser:
Do not merge this yet when #208 or a variation of gets merged I will update this to use the eg.WindowsVersion

kdschlosser:
OK I have made the changes to support the new eg.WindowsVersion. so everything should be good to go on this...  I have been using the changes to how the messages are registered for some while without any errors.. and I know that @per1234 has used this as well... I have not heard a report back.. so I am going to take it as no news is good news.

plugins/System/PowerBroadcastNotifier.py
```diff
@@ -62,10 +62,15 @@
 AWY_EXITING = 0x0
 AWY_ENTERING = 0x1
 
-if WIN_7:
-    monGUID = '{02731015-4510-4526-99e6-e5a17ebd1aea}'
+
+if eg.WindowsVersion >= 8:
```
topic2k: needs to be: `if eg.WindowsVersion() >= 8:`

plugins/System/PowerBroadcastNotifier.py
```diff
@@ -172,38 +188,40 @@ class POWERSCHEME_PERSONALITY(ctypes.Structure):
...
     def __init__(self, plugin):
         self.plugin = plugin
+        self.monitorNotify = None
+        self.awayNotify = None
+        self.sourceNotify = None
+        self.batteryNotify = None
+        self.savingNotify = None
+        self.schemeNotify = None
 
+        if eg.WindowsVersion >= 'Vista':
```
topic2k: needs to be `if eg.WindowsVersion() >= 'Vista':`

plugins/System/PowerBroadcastNotifier.py
```diff
@@ -172,38 +188,40 @@ class POWERSCHEME_PERSONALITY(ctypes.Structure):
...
-        )
-
     def Close(self):
-        Unregister(self.monitorNotify)
-        Unregister(self.awayNotify)
-        Unregister(self.sourceNotify)
-        Unregister(self.batteryNotify)
-        Unregister(self.savingNotify)
-        Unregister(self.schemeNotify)
+        if eg.WindowsVersion >= 'Vista':
```
topic2k: needs to be `if eg.WindowsVersion() >= 'Vista':`

kdschlosser:
see #211

kdschlosser:
I made the changes... hopefully this is good to go.

PR #208, opened 2017-03-03
[EventGhost] - Enhancement - Windows version checking

kdschlosser:
Adds a simple way to grab the windows version. 

eg.WindowsVersion.IsXp()
eg.WindowsVersion.IsVista()
eg.WindowsVersion.Is7()
eg.WindowsVersion.is8()
eg.WindowsVersion.Is10()

This is going to be used by the System.PowerBroadcastNotifier (when modified to do so, due to a bug) and also System.DeviceChangeNotifier (if the PR gets accepted)

### DIFF
--- a/eg/Classes/WindowsVersion.py
+++ b/eg/Classes/WindowsVersion.py
@@ -1,0 +1,110 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of EventGhost.
+# Copyright © 2005-2016 EventGhost Project <http://www.eventghost.org/>
+#
+# EventGhost is free software: you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the Free
+# Software Foundation, either version 2 of the License, or (at your option)
+# any later version.
+#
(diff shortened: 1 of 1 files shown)


topic2k:
Nice. What about turning it into a class and add compare functions? `eg.WindowsVersion >= "Vista"`
[See here](https://gist.github.com/topic2k/254774adbafffd73204934e9211009ce)

kdschlosser:
OK ya out did me on this one. LOL.. your is better. I never thought about adding comparison.

Only one thing i would add to it tho..

    class WindowsVersionError(Exception):

        def __init__(self, msg):
            self.msg = msg

        def __str__(self):
            return 'Windows %s is not supported.' % self.msg 

and change out that ValueError to call this. I think we need to start providing a little better traceback information. ValueError to a newish coder/eg user is probably going to mean nothing to them. but this will..

and in all honesty it can help us a lot as well. but go ahead. you rewrote my whole thing. i get credit for the basic idea and you get the credit for the code.. so tank this PR and do a PR for yours

PR #211, opened 2017-03-11
Fixes incorrect reporting of windows version

kdschlosser:
Initializes the class so every time you want to use the comparison feature you don't have to.


topic2k:
I had that already included, but then the generating of the helpfile for this class doesn't work.

kdschlosser:
that's strange that the docs don't build right if it is initialized...

but anyways there was a glitch in it... all() doesn't report properly and also when grabbing the versions from the dict you were turning a list of lists into a list of lists of lists which a single iteration over is not going to work properly. XP and 8 have such a case. I also fixed the versions for windows 8 and windows 7. windows 7 only has 1 version and windows 8 has 2

topic2k:
Good findings. Now that i look at it again, i see my errors. Instead of all() i should have used any(). Can you please try it with `any()` instead of `bool(max(`. I think that would look a little cleaner (pythonic).

kdschlosser:
it works with any()

kdschlosser:
and No worries. late nights with little sleep does that to me as well...

PR #213, opened 2017-03-17
[System] - Enhancement - Adds enabling or disabling of network adapters

kdschlosser:
The title says it all.

### DIFF
--- a/plugins/System/__init__.py
+++ b/plugins/System/__init__.py
@@ -138,6 +138,9 @@ class Text:
     class SoundGroup:
         name = description = "Audio"
 
+    class NetworkGroup:
+        name = description = "Network"
+
     forced        = "Forced: %s"
     forcedCB      = "Force close of all programs"
     primaryDevice = "Primary Sound Driver"
--- a/plugins/System/__init__.py
+++ b/plugins/System/__init__.py
@@ -26,6 +26,7 @@ import wx
 import _winreg
 import win32con
 import win32gui
+import win32com
 from base64 import b64decode, b64encode
 from PIL import Image
 from qrcode import QRCode, constants as QRconstants
@@ -72,8 +73,9 @@ eg.RegisterPlugin(
     author = (
         "Bitmonster",
         "blackwind",
+        "K"
     ),
-    version = "1.1.10",
+    version = "1.1.11",
     description = (
         "Actions to control various aspects of your system, including "
         "audio, display, power, and registry."
@@ -215,6 +217,15 @@ class System(eg.PluginBase):
         )
         group.AddAction(Registry.RegistryChange)
         group.AddAction(Registry.RegistryQuery)
+
+        if eg.WindowsVersion() >= 'Vista':
+
+            group = self.AddGroup(
+                text.NetworkGroup.name,
+                text.NetworkGroup.description,
+            )
+            group.AddAction(EnableNetworkAdapter)
+            group.AddAction(DisableNetworkAdapter)
 
     def __start__(self):
         eg.Bind("ClipboardChange", self.OnClipboardChange)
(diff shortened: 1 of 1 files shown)


kdschlosser:
Nothing happens?? That's very strange. It works on my machine. The adapter will only get enabled or disabled if it is actually installed in the computer. I tried to filter it so it would only show adapters that were actually installed. And it will only disable an adapter that is enabled. And vice versa. And I have to look at it again but I believe it also has to be a good standing device. One without an error. I will retest it. 

#200 has an action that will return a device so if you wanted to get the status of a network card that would be the way to go. But I have been thinking of making this Device thing into its own core plugin. The System plugin is getting a tad bloated and hard to follow for my liking. And I think it needs to be split up a bit. And then I can make a bucket of convince actions for the various devices. I was thinking about making actions for the following device types. Hard disks, Removable disks, Network cards, BlueTeeth devices

topic2k:
Looks like `Win32_NetworkAdapter` is [deprecated](https://msdn.microsoft.com/en-us/library/aa394216(v=vs.85).aspx) and you need to use [MSFT_NetAdapter](https://msdn.microsoft.com/en-us/library/hh968170(v=vs.85).aspx) for Windows 8 and later.

```python
import win32com
wmi = win32com.client.GetObject("winmgmts:\\root\\StandardCimv2")
query = 'Select * from MSFT_NetAdapter'
adapters = wmi.ExecQuery(query)
for device in adapters:
    print device.DeviceID, '|', device.Name, '|', device.DriverDescription, '|', device.InterfaceDescription, '|', device.InterfaceName
```
results for my machine:
```
{A07051E1-F148-4B24-B8AC-BE359A9542DB} | Hobbingen-LAN | Realtek PCIe GBE Family Controller | Realtek PCIe GBE Family Controller #2 | ethernet_32770
{13BFA5C1-E1C8-472B-A5F8-E1C3CF5DCACD} | WLAN | Intel(R) Centrino(R) Wireless-N 2230 Driver | Intel(R) Centrino(R) Wireless-N 2230 | wireless_32768
{69A3FD7E-3740-46FA-9A5D-DECC1B67C060} | Ethernet 4 | LANCOM Secure Client Virtual NDIS6 Adapter | LANCOM Secure Client Virtual NDIS6 Adapter | ethernet_32774
{38195D21-B912-4DFF-A782-AA93AE2E83A1} | VirtualBox Host-Only Network | VirtualBox Host-Only Ethernet Adapter | VirtualBox Host-Only Ethernet Adapter | ethernet_32775
{24EF7112-37A0-4391-8F11-34216AB8F906} | Ethernet | Realtek PCIe GBE Family Controller | Realtek PCIe GBE Family Controller | ethernet_32768
```

But when i tried to disable my network-adapter, i get a com_error: access denied.
```python
result = wmi.ExecQuery(query+ ' WHERE InterfaceDescription="%s"' % 'Realtek PCIe GBE Family Controller')
lan = result[0]
lan.Disable()
```
```
  File "<input>", line 1, in <module>
  File "C:\pyvenv\EG.org_py27sl\lib\site-packages\win32com\client\dynamic.py", line 516, in __getattr__
    ret = self._oleobj_.Invoke(retEntry.dispid,0,invoke_type,1)
com_error: (-2147352567, 'Ausnahmefehler aufgetreten.', (0, u'SWbemObjectEx', u'Zugriff verweigert. ', None, 0, -2147217405), None)
```

kdschlosser:
try this class

    MSFT_NetAdapterPowerManagementSettingData

kdschlosser:
you will have to change the WQL to be 

    WHERE Name=

topic2k:
I retried again. This time i run EG as Administrator. On Win7 i can then disable/enable a network adapter. On Win10 i can disable a adapter. But when i want to enable it, i get an exception that the adapter wasn't found.
Then i checked again with `MSFT_NetAdapter`. That worked. I pushed it to my [fork](https://github.com/topic2k/EventGhost/tree/enhancement-system-networkadapters) so can [look](https://github.com/kdschlosser/EventGhost/compare/enhancement-system-networkadapters...topic2k:enhancement-system-networkadapters) at it.

kdschlosser:
in windows you do have to have admin privileges to enable or disable devices...  that is most likley the root cause of it not working.. but is there a way to programmatically cause an admin login? if so i could provide a place to key in the admin username and password

kdschlosser:
maybe possibly provide a place to enter this globally instead of on an action by action basis would be a good idea also.. so if it's globally entered then it won't ask. but if not it has to be done on an action by action basis.

kdschlosser:
https://msdn.microsoft.com/en-us/library/aa826699%28v=vs.85%29.aspx

this tells you what is filtered by the UAC with WMI

kdschlosser:
I am going to remove this only because the only way to achieve this is to either edit the security descriptors for WMI to allow the user to run WMI methods or to run EG as administratr. and neither of these ways is nice.

PR #210, opened 2017-03-05
Add class for windows version checking

topic2k:
replaces #208 
based on the idea from @kdschlosser, extended to a class with compare functions.

Check for a specific Windows version:
`eg.WindowsVersion.IsVista()`  or `eg.WindowsVersion == "Vista"`

Check for a minimum Windows version:
`eg.WindowsVersion >= "Vista"`

**Notice for plugin developers:** `eg.Utils.IsXP()` and `eg.Utils.IsVista()` are now deprecated.

---
**Functions:**
```
IsXP()
IsVista()
Is7()
Is8()
Is80()
Is81()
Is10()
```
**Comparison:**
```
eg.WindowsVersion [<|<=|==|!=|>=|>] "[XP|Vista|7|8|80|81|10]"

XP:    Windows XP
       Windows Server 2003
       Windows 2003 R2
       Windows Tablet PC
       Windows Media Center Edition 2002
       Windows Media Center Edition 2004
       Windows Media Center Edition 2005
Vista: Windows Vista
       Windows Server 2008
7:     Windows 7
       Windows Server 2008 R2
8:     (80 and 81)
80:    Windows 8
       Windows Server 2012
       Windows RT
81:    Windows 8.1
       Windows 2012 R2
       Windows RT 8.1
10:    Windows 10
       Windows Server 2016
```

PR #215, opened 2017-03-24
[Eventghost] - Enhancement - WindowsVersion, Fixes issue with help

kdschlosser:
Fixes issue with help not being generated if WindowsVersion gets constructed. 

I used Cli.args.isMain to determine if the build process is running. this allows for eg.WindowsVersion to not be constructed and the help to be built


topic2k:
just for reference: #210

kdschlosser:
No worries m8. Knew we would find a way around that help problem. Took a bit to think of it.

PR #221, opened 2017-04-20
Fixes incorrect replacement usage of IsVista() and IsXP()

kdschlosser:
The original Utils.IsVista() and Utils.IsXp()  would return True ot False if the OS was greater than or eqiual to. We had changed everything to use the new WindowsVersion.IsVista() and IsXP() which would only return True/False if it is that opperating system.

This fixes all the wrongly used IsVista() and IsXP() to use WindowsVersion >= 'Vista' and WindowsVersion >= 'XP'

### DIFF
--- a/eg/Utils.py
+++ b/eg/Utils.py
@@ -394,11 +394,11 @@ def IsVista():
     """
     warnings.warn(
         "eg.Utils.IsVista() is deprecated. "
-        "Use eg.WindowsVersion.IsVista() instead",
+        "Use eg.WindowsVersion >= 'Vista' instead",
         DeprecationWarning,
         stacklevel=2
     )
-    return eg.WindowsVersion.IsVista()
+    return eg.WindowsVersion >= 'Vista'
 
 def IsXP():
     """
@@ -406,11 +406,11 @@ def IsXP():
     """
     warnings.warn(
         "eg.Utils.IsXP() is deprecated. "
-        "Use eg.WindowsVersion.IsXP() instead",
+        "Use eg.WindowsVersion >= 'XP' instead",
         DeprecationWarning,
         stacklevel=2
     )
-    return eg.WindowsVersion.IsXP()
+    return eg.WindowsVersion >= 'XP'
 
 def LogIt(func):
     """
--- a/eg/Classes/App.py
+++ b/eg/Classes/App.py
@@ -25,7 +25,7 @@ import wx
 import eg
 from eg.WinApi.Dynamic import ExitProcess, SetProcessShutdownParameters
 
-IS_VISTA = eg.WindowsVersion.IsVista()
+IS_VISTA = eg.WindowsVersion >= 'Vista'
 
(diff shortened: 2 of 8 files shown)


kdschlosser:
This is kind of a biggie, and is going to have a huge impact on how EG runs. I would push out another release candidate ASAP. My opinion...  ¯\_(ツ)_/¯

eg/Utils.py
```diff
@@ -394,23 +394,23 @@ def IsVista():
     """
     warnings.warn(
         "eg.Utils.IsVista() is deprecated. "
-        "Use eg.WindowsVersion.IsVista() instead",
+        "Use eg.WindowsVersion >= 'Vista' instead",
```
topic2k: `IsVista()` means *is Vista* and not *is Vista or later*. So i think the message can/should be changed to `"Use eg.WindowsVersion.IsVista() or eg.WindowsVersion >= 'Vista' instead."`

eg/Utils.py
```diff
@@ -394,23 +394,23 @@ def IsVista():
...
+    return eg.WindowsVersion >= 'Vista'
 
 def IsXP():
     """
     Determine if we're running XP or higher.
     """
     warnings.warn(
         "eg.Utils.IsXP() is deprecated. "
-        "Use eg.WindowsVersion.IsXP() instead",
+        "Use eg.WindowsVersion >= 'XP' instead",
```
topic2k: Same as with IsVista 😉

kdschlosser: well the original eg.Utils.IsVista() and eg.Utils.IsXP() would return true if the os was >= vista or xp. hence the reason why i only did the >= so this way it will mimic the exact operation of the original functions. where as the new version only returns  true if it is that exact os.

if you look at the original code
    
    def IsVista():
        """
        Determine if we're running Vista or higher.
        """
        return (sys.getwindowsversion()[0] >= 6)

    def IsXP():
        """
        Determine if we're running XP or higher.
        """
        return (sys.getwindowsversion()[0:2] >= (5, 1))

even the doc string says and higher,


I'll do whatever it is you want me to do on this. but i just wanted to have to depreciation warning show exactly how to use the new to replace the old. no confusion that way

kdschlosser: let me know what to do ob-wan

topic2k: The return value of `eg.Utils.IsVista()` is correct. I only meant the deprecation message. Just to make sure that the user/developer notice that there is a difference behaviour of `eg.Utils.IsVista()` and `eg.WindowsVersion.IsVista()`.
But now while i write this, i think you are right and only display `>= 'Vista'` because that's how Utils.IsVista() worked.
 I will merge it now and make a new release.

kdschlosser: cool. Thanks for the quick update on this. I am not sure if you want to change up the forum RC1 to inform the users not to use this version that is a large flaw in it. but leave it there for history purposes.

Issue #226, opened 2017-05-02
MainMessageReceiver problem with ChangeClipboardChain


saue0:
The ChangeClipboardChain do not seem to be call.
  The workaround was Removing it from Stop into Finish.

Can someone verifie this.

Thanks

kdschlosser:
I have no idea what you are having an issue with.

I need errors, a better description of the problem. screen shots. I need something other then a very vague description to be able to help ya out. My best guess is that you removed it from MainMessageReceiver.py but you do not explain what the problem was

saue0:
The problem is that evenghost do not unregister it self from the clipboard chain , it's like if the stop procedure do not get call. By moving the  ChangeClipboardChain in the finish procedure, eventghost get unregisted. This poblem cause other program that capture the clipboard change (ex Ditto) to stop receiving notification after eventghost is stopped, The chain is break.

kdschlosser:
give a way to replicate this problem using a common program like notepad or wordpad

kdschlosser:
I downloaded that clipboard program Ditto that you mentioned.. I copied some text and it appeared in Ditto. then i started EG and copied some text and it appeared in Ditto. then i closed EG and copied some text and it appeared in Ditto. So i am not sure exactly what the problem is.. I am thinking that is what you meant. but everything seems to work just fine.

kdschlosser:
it could have something to do with you OS??... what version of windows are you running? I tested this on a Windows 7 computer. It could be a Windows 10 thing if that is what you are running.

saue0:
I'm using windows 10, I will retry monday with the latest version of EG. 
How can I add log info in the stop procedure ? 
"print" do not work because EG is stopping.  need to log in a file.

kdschlosser:
it is being logged there is a decorator @eg.LogIt before the action that gets called is there is a clipboard chain change. 

so what is supposed to happen is when EG gets closed the Stop() method gets called and the ChangeClipboardChange command gets called to Change the chain. and when that happens windows is supposed to send out a notification to all windows that are set to receive these notification and the callback is OnChangeClipboardChain. which should be logged in the debugging log. which can be turned on if you start EG from a commandline with the -debug switch and the file is located in %appdata%\EventGhost as Log.txt

 trigger a windows event

saue0:
Thanks will try and get back soon.

kdschlosser:
any new information on this?

saue0:
[...]

PR #227, opened 2017-05-09
MainMessageReceiver: Fixes broken clipboard chain in windows 10 (reported by saue0)

kdschlosser:
Windows 10 is a real fickle about having the same thread that registers for notifications also remove the notifications. in the case of the MainMesageReceiver it handled the ChangeClipboardChain which would remove EG from receiving clipboard notifications. if the process is not done right and the chain doesn't get repaired when EG closes other programs will stop receiving notifications that there is new or changed clipboard data. This moves the call to change the chain into the MainMessageReceiver thread so the clipboard chain get properly repaired when EG closes


kdschlosser:
resolves #226 
please do me a favor and @saue0 gets credit for this.

Issue #229, opened 2017-05-21
egplugin import doesn't work if EG is executed as admin


ThomasBott:
If EventGhost is executed and running as administrator and you double click an .egplugin file, a new EG process is created, but you don't see any UI for it. (so for a user nothing happens)
If you now close EG and wait some time, EG comes up again with a warning "Failed to launch for unknown reasons".

EG 0.5 RC 2

kdschlosser:
ok cool. I know what is causing that. it's going to take a bit to figure out a way to fix it.

kdschlosser:
"Failed to launch for unknown reasons". there is a number associated with this error. do you know what it is?

ThomasBott:
Yes, it's -2146959355

kdschlosser:
TY that's the number i wanted

kdschlosser:
here is a link for a build i just did up. I am running windows 7. I want to make sure it work properly on windows 10. If you are interested in testing it out for me, it is the same thing as rc2 except for this change. but here is the link.. I have also attached the file with the fix as well just in case you do not want to run the installer you can simply backup the Cli.py file and put the new one in it's place. It appears as tho whoever added the error checking for the elevated permissions used the wrong error codes. this has been fixed. but I want to test if we can completely do away with any kind of a message box and still get the plugin to install.

Now it may launch another EG window to accomplish this. and if it does let me know. because the plugin list in the first EG would not be refreshed. So I would need to put into place a mechanism to send the path to the plugin file to the EG that has the elevated permissions. But this is a place to start. On windows 7 it currently opens up the plugin install dialog. I just want to cover all the bases with this

here is the link

https://ci.appveyor.com/project/kdschlosser/eventghost-9ivlx/build/366-bugfix-plugin-install-elevated/artifacts

and the Cli.py file is attached. I had to change the extension to .txt so you will need to change it to .py

[Cli.txt](https://github.com/EventGhost/EventGhost/files/1020564/Cli.txt)

kdschlosser:
Any update on this? Has it been tested? If it works I would like to get a PR issued for it

topic2k:
Seems to not work on Win10-64bit. I started EG as administrator and then double clicked as user on xxx.egplugin file. In task manager there come up two more processes, but nothing happens.

kdschlosser:
blah!

this is going to be a tough nut to crack.

kdschlosser:
[...]

PR #240, opened 2017-06-12
Add Named pipe support to solve the problem with elevations

kdschlosser:
If the first running instance of EG is elevated and the second is not windows com throws an error. this will restart EG as elevated if this error occurs.

### DIFF
--- a/eg/Classes/App.py
+++ b/eg/Classes/App.py
@@ -133,7 +133,6 @@ class App(wx.App):
             eg.PrintDebugNotice("The following threads did not terminate:")
             for thread in threads:
                 eg.PrintDebugNotice(" ", thread, thread.getName())
-
         eg.PrintDebugNotice("Done!")
         ExitProcess(0)
 
--- a/eg/Classes/App.py
+++ b/eg/Classes/App.py
@@ -134,7 +134,6 @@ class App(wx.App):
             for thread in threads:
                 eg.PrintDebugNotice(" ", thread, thread.getName())
 
-        eg.namedPipe.stop()
         eg.PrintDebugNotice("Done!")
         ExitProcess(0)
 
--- a/eg/Classes/App.py
+++ b/eg/Classes/App.py
@@ -133,6 +133,8 @@ class App(wx.App):
             eg.PrintDebugNotice("The following threads did not terminate:")
             for thread in threads:
                 eg.PrintDebugNotice(" ", thread, thread.getName())
+
+        eg.namedPipe.stop()
         eg.PrintDebugNotice("Done!")
         ExitProcess(0)
 
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -146,6 +146,7 @@ install:
       } Else {
         # WIP
         python "_build\Build.py" --build --package
+        eventghost -install
       }
       $Env:SetupExe = gci -recurse -filter "_build\output\*Setup.exe" -name
       " "
--- a/_build/data/InnoSetup.template
+++ b/_build/data/InnoSetup.template
@@ -205,7 +205,7 @@ Root: HKCR; Subkey: "EventGhost Plugin\DefaultIcon"; ValueType: string; ValueNam
 Root: HKCR; Subkey: "EventGhost Plugin\shell\open\command"; ValueType: string; ValueName: ""; ValueData: """{app}\EventGhost.exe"" ""%%1"""
 
 [Run]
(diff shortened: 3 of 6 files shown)


kdschlosser:
I am not sure as to why appveyor fails. but when I build it builds fine.

kdschlosser:
please make sure you read the commit comments on the last commit. as i feel this addition can bring some nifty things to EG 

here is some sample code if you want to tell EG to do different things from another program. the code example you run from a python interpreter outside of EG. and this will cause EG to minimize.

    import win32pipe
    import win32file

    def send_message(msg):
        p = win32pipe.CreateNamedPipe(
            r'\\.\pipe\eventghost_pipe',
            win32pipe.PIPE_ACCESS_DUPLEX,
            win32pipe.PIPE_TYPE_MESSAGE | win32pipe.PIPE_WAIT,
            1,
            65536,
            65536,
            300,
            None
        )

        win32pipe.ConnectNamedPipe(p, None)
        win32file.WriteFile(p, msg)

    send_message('eg.document.frame.Iconize, (True,)')

topic2k:
still no luck.
```python
  File "EventGhost.pyw", line 53, in <module>
  File "c:\Program Files (x86)\EventGhost\eg\__init__.py", line 31, in <module>
    import Cli
  File "c:\Program Files (x86)\EventGhost\eg\Cli.py", line 146, in <module>
    args.pluginFile
  File "c:\Program Files (x86)\EventGhost\eg\NamedPipe.py", line 164, in send_message
    None,
pywintypes.error: (5, 'CreateFile', 'Access denied')
```
(EG running as Administrator and normal user doubleclicked on a .egplugin file)

kdschlosser:
I also want you to run EG s a normal user and install a plugin as Administrator to see if this will handle that scenario and well. I am thinking it will.

kdschlosser:
I also wanted to add this as well because i forgot to

Resolves #229

kdschlosser:
One other not on this.... !!! YAY!!!! this is going to help out a HUGE amount if we manage to get this going. 

But check this out. You know how if you go to install EG and you have an instance already and it asks you to close it?. well this is the mechanism to do it automatically.

But i mention this because of #245 and what exactly is getting stuck in the builder i am not sure. I am thinking it is when eg gets run with -install to create the .pyc files. I don't specifically know what the reasoning is behind launching EG to compile those files. because python does this when the file (module) is imported, it doesn't speed anything up.

@topic2k do you know what the purpose to it is?

topic2k:
> @topic2k do you know what the purpose to it is?

no, not really. I can only guess that it was to speed up (first) loading of EG. Maybe in older python version it was a speed improvement.
Secondly it creates a COM server. Could it be needed to have EG be registered to open .egplugin/.egtree?

What if you comment that line in the setup template? (line 208, put a ; in front) Add as commit here and i will test.

kdschlosser:
Fantastic! No more hang when installing EG and the user is in the Administrators group.. woo hoo

kdschlosser:
Everything loads up just fine. there is a very slight performance hit the first time it loads up. but it's really really tiny.

kdschlosser:
all of the COM server methods/functions can get depreciated if we get this named pipe thing going properly. they will become obsolete.

kdschlosser:
I got it working!

eg/Classes/App.py
```diff
@@ -133,6 +133,7 @@ def OnExit(self):
             eg.PrintDebugNotice("The following threads did not terminate:")
             for thread in threads:
                 eg.PrintDebugNotice(" ", thread, thread.getName())
+
```
topic2k: I think this has nothing to do with the PR and should be left out or made in another PR for code cleanup. Same goes for other whitespace/blank line fixes

topic2k:
fixes #245, fixes #229

topic2k:
Let's say i'm running EG with full elevation. Now from a command line with limited elevation i call `eventghost.exe -restart`. EG will be restarted, but now with limited elevation. Can we check what elevation the running EG has and automatically decide if we restart with full or limited elevation?

kdschlosser:
run eventghost -restart with elevated privlages

eg/Classes/App.py
kdschlosser: HA sorry about that. that happened because i didn't have the thread that runs the named pipe set up as a daemon thread at first and that is where i told the thread to stop. i forgot to remove the extra line. my bad

PR #241, opened 2017-06-12
Fixes incorrect checking of powerbroadcast notification message

kdschlosser:
In an iteration loop to check the notification message an exception could get raised. this adds handling of that exception.


kdschlosser:
@per1234 this could be the solution to your monitor issue. You should give it a go.

PR #243, opened 2017-06-12
[USBUIRT] Fixes learn ir dialog not closing

kdschlosser:
I have no means to test this. but when looking at the changes to the MCERemoteVista plugin that has this same issue I instituted the same changes.


topic2k:
for reference: http://www.eventghost.net/forum/viewtopic.php?f=4&t=9698&p=46540&hilit=USBUIRT#p46342

PR #246, opened 2017-06-24
[System] fix GetMasterVolume (move use of VistaVolEvents to eg.WinApi.SoundMixer)

kdschlosser:
I have had a person ask me why the eg.WinApi.SoundMixer.GetMasterVolume is not functioning. and after a bit of digging through the code I discovered the reason for this. It has to do with VistaVolEvents only being used inside the system plugin and the person running a version of windows that is greater then Vista. So I have moved the use of VistaVolEvents to eg.WinApi.SoundMixer.

This has gotten rid of the voodoo code to override the __call__ methods of the sound actions to support the use of vista vol events. But this will also make it easier to change the package/API that is being used to control the sound devices. and it makes the functionality of SoundMixer work across all Windows versions.


topic2k:
for reference: http://www.eventghost.net/forum/viewtopic.php?f=2&t=9666#p45933

kdschlosser:
I am hoping you tested this thing. i was running into an issue with it causing my sound card to crash. but I sometimes have problems with EG generating sound card events and I have to reboot the computer then it starts working again. this happens when i unplug the speakers. and I didn't reboot to see if that is what it was. i figured you would have told me if it wasn't working properly

topic2k:
At least here on Win10 i get events when changing volume and your test script shows me values for my six audio devices. OTOH my speaker is the one from my monitor through HDMI, so i can't unplug.

```python
for i in range(10):
    try:
        volume = eg.WinApi.SoundMixer.GetMasterVolume(i)
        print 'deviceid:', str(i), ' Volume:', str(volume)
    except:
        pass
```

kdschlosser:
what i am saying is even the original setup had the issue with the speakers. this is not a new problem. it's in the vistavolevents it's self. just that i didn't test my changes all of the way because i didn't reboot the computer to fix the issue that i was having. so i wanted to make sure you ran the system plugin actions like the mote and the setting volume. things like that

topic2k:
Ah ok, you're right. If i use the default example egtree and change the volume control macros to use device "Primary Sound Driver", then i get SoundMixerExceptions.

kdschlosser:
You will always get those errors I believe. Even before the change I made.

PR #247, opened 2017-06-26
NamedPipe: Fixes possible endless loop if -restart is used at the CLI

kdschlosser:
Fixes the possibility of getting into an endless loop if the eg instance needs to be saved and -restart has been used

This change also added the ability to send back through the pipe any returned data

### DIFF
--- a/eg/Cli.py
+++ b/eg/Cli.py
@@ -118,18 +118,45 @@ if args.isMain:
             import NamedPipe
 
             if args.restart:
-                import time
+                # import time
 
-                NamedPipe.send_message('eg.app.Exit, ()')
+                try:
+                    if NamedPipe.send_message('eg.document.IsDirty, ()'):
+                        answer = ctypes.windll.user32.MessageBoxA(
+                            0,
+                            'EventGhost cannot restart.        \n\n'
+                            'Save configuration changes?.      \n',
+                            "EventGhost Restart Error",
+                            3 | 40000
+                        )
 
-                while ctypes.GetLastError() != 0:
-                    time.sleep(0.1)
+                        if answer == 2:
+                            ctypes.windll.kernel32.ExitProcess(0)
+                        elif answer == 7:
+                            NamedPipe.send_message(
+                                'eg.document.SetIsDirty, (False,)'
+                            )
+                        elif answer == 6:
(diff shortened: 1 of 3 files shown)


eg/Cli.py
```diff
@@ -118,18 +118,45 @@ class args:
...
-                        "Global\\EventGhost:7EB106DC-468D-4345-9CFE-B0021039114B"
-                    )
+                # import time
+
+                try:
+                    if NamedPipe.send_message('eg.document.IsDirty, ()'):
+                        answer = ctypes.windll.user32.MessageBoxA(
+                            0,
+                            'EventGhost cannot restart.        \n\n'
+                            'Save configuration changes?.      \n',
```
topic2k: I think we should take the same wording as when normally closing EG with unsaved changes:
```
"Configuration contains unsaved changed.\n\n"
"Do you want to save before continuing?"
```

kdschlosser: well you are correct sir. TY for noticing :P

kdschlosser: HA!!  I thought my wording was wrong and you go and lay something on me that is worse off....

    "Configuration contains unsaved changed.\n\n"
    "Do you want to save before continuing?"

LMAO

I will fix the wording in document.py and issue a PR for it. and I will also change the wording to match in this PR

kdschlosser:
Just wanted to let ya know I made the requested changes

eg/Cli.py
```diff
@@ -149,7 +150,8 @@ class args:
                     if not NamedPipe.send_message('eg.app.Exit, ()'):
                         ctypes.windll.user32.MessageBoxA(
                             0,
-                            'EventGhost cannot restart - unknown error.\n',
+                            'EventGhost cannot restart.             \n\n'
+                            'Unknown Error.                         \n',
```
topic2k: One more question: will we get any exception or error number that we could give as addidtional information?

kdschlosser: Nope. but in all reality that error should never occur. if it does then there is something that simply does not work right. and I would have no clue what it could be. because the only reason why eg.app.Exit() would return anything but true would be if the user canceled the saving of the changes from a dialog that should not have appeared in the first place,

PR #249, opened 2017-06-27
[EventGhost] - Enhancement - Log.py, Upgrades the debug logging

kdschlosser:
Changes the debug logging to find the name of the calling module and if it is a plugin it will create a new file specifically for that plugin. Provides alittle more organization to the debug logging. It also houses all of the log files into a folder called logs in %appdata%\EventGhost. 

It also fixes the issue of not having the debug header information added if the user has logging turned on from inside of the GUI. and I added the Windows version, Windows major, minor, build version, the start time of EG is all added to this header also. The header is added to each new file that is made or appended to one that is already existing. It closes the file after each write to it so it can be deleted without the need to shutdown EG.

If there happen to be any error during the load of EG this information will be written to the standard Log.txt file. and anything that is not a plugin that is writing log information is written to Core.txt in the logs folder.

### DIFF
--- a/eg/Classes/Log.py
+++ b/eg/Classes/Log.py
@@ -19,6 +19,7 @@
 import codecs
 import sys
 import wx
+import os
 import inspect
 import platform
 from collections import deque
@@ -120,9 +121,7 @@ class Log(object):
             def write(self, data):
                 log.Write(data, ERROR_ICON)
                 if eg.debugLevel:
-                    splitLog = getattr(eg.config, 'splitDebuggingLog', None)
-
-                    if splitLog:
+                    if getattr(eg.config, 'splitDebuggingLog', None):
                         module_name = caller_name()
 
                         if 'CorePluginModule' in module_name:
@@ -134,18 +133,9 @@ class Log(object):
                             module_name = 'Core'
 
                         if module_name not in debug_logs:
-                            import os
-
-                            prgName = os.path.splitext(
-                                os.path.basename(sys.executable)
-                            )[0]
-                            prgAppDataPath = os.path.join(
-                                os.environ["APPDATA"],
-                                prgName
-                            )
 
                             logPath = os.path.join(
-                                prgAppDataPath,
+                                eg.configDir,
                                 "logs"
                             )
                             logFilePath = os.path.join(
@@ -170,12 +160,10 @@ class Log(object):
                             f,
                             'backslashreplace'
                         )
+                        close = f.close
                     else:
-
-                        class f:
-                            @staticmethod
-                            def close():
-                                pass
+                        def close():
+                            pass
 
                         file_writer = sys.stderr
                         if not self.header_printed:
@@ -187,7 +175,7 @@ class Log(object):
                     except:
                         file_writer.write(data.decode("mbcs"))
                     finally:
-                        f.close()
+                        close()
                 else:
                     if self.header_printed:
                         self.header_printed = False
--- a/eg/Classes/Log.py
+++ b/eg/Classes/Log.py
@@ -19,6 +19,8 @@
 import codecs
 import sys
 import wx
+import inspect
+import platform
 from collections import deque
 from threading import currentThread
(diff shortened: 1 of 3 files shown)


eg/Classes/Log.py
```diff
@@ -34,11 +36,59 @@
...
+    strftime('      %d/%m/%Y - %H:%M:%S'),
+    '-' * 40,
+    '%s Version: %s' % (
+        eg.APP_NAME,
+        eg.Version.string
+    ),
+    'Windows Version: %s' % ' '.join(
+        list(platform.win32_ver())
+    ),
+    'Computer Name: %s' % platform.node(),
```
topic2k: Do we really need the computer name for debugging?

eg/Classes/Log.py
```diff
@@ -34,11 +36,59 @@
...
+    'Processor: %s' % platform.processor(),
+    'Python: %s %s' % (
+        platform.python_version(),
+        platform.python_implementation()
+    ),
+    'Python Compiler: %s' % platform.python_compiler(),
+    'Python Build Date: %s' % (
+        platform.python_build()[1],
```
topic2k: I'm missing the info that it's stackless python, so i would just use `sys.version`. It returns `'2.7.12 Stackless 3.1b3 060516 (default, Oct 31 2016, 18:42:08) [MSC v.1500 32 bit (Intel)]'
`

topic2k (commented):
I personaly don't like to have the log splitted, but that's  just me. Don't know what others say. I hope we get some more comments.

eg/Classes/Log.py
kdschlosser: ok can do
I'll do it right now

kdschlosser: do you like the way it works tho? I likle it alot better. it is a pain in the backside to sift through a debug log file when you are looking for something that is specific to a plugin. This is a nice way to handle it.

topic2k: maybe if i would use EG more then i would find it more usefull :D but i will merge it anyway ;)

kdschlosser: all done.

kdschlosser:
corrected the python version.
removed the computer name.

added the ability to turn on or off the slpit logging via adding a line to the config.py file

    splitDebuggingLog = True

made it so that each time the debugging log is turned on or off it will set a flag for printing the header to the original Log.txt

kdschlosser:
Oh and this will only split off the plugins and nothing else. I am not usually one for split logs either. But when i specifically am looking for output form a plugin i am working on it's an annoyance to have to go reading through a large log and this way I am also able to only delete that plugins log.

topic2k:
Thanks for adding the split option 😃 
Anything else you want to add, or is it ready to merge?

kdschlosser:
I do not think anything else needs to be added. just want to make sure it works 100% because it does have some voodoo code to it. and that is not my favorite kind of code. but it's the only way to do it without adding a method to the plugin base. and then having the plugin devs update their plugins to support it. so i though this would be a better way

topic2k:
Now EG only works if i have `splitDebuggingLog = True`. If i set it to `False` or omit it, then EG hangs. After some trying it seems that the logging crashes if the `logs` directory doesn't exists.

kdschlosser:
i just wanted to let you know this one should be fixed. i forgot to mention it. i did a commit a while back

topic2k:
well, sorry to say no. I've deleted Appdata\EventGhost and then run EG with -debug option. The GUI doesn't show up, the process seems to hang and a log file with only `File "C:\Users\topic\AppData\Roaming\EventGhost\config.py" does not exist.` as content is created. After while the EG process exits.
If i run EG without debug option it starts normally.

kdschlosser:
after it runs normally if you then close EG and restart it with the -debug parameter does it work then?

topic2k:
No, makes no different if i run it successfully before using -debug.

PR #253, opened 2017-06-29
Moves initialization of python paths to the first thing that EG does

kdschlosser:
By moving the initialization of the python paths to the very startup of EG this allows for us to be able to import WinApi modules early on in the startup process without encountering an import exception for cFunctions

### DIFF
--- a/eg/Classes/FolderPath.py
+++ b/eg/Classes/FolderPath.py
@@ -16,8 +16,11 @@
 # You should have received a copy of the GNU General Public License along
 # with EventGhost. If not, see <http://www.gnu.org/licenses/>.
 
+from os.path import join, split
 from ctypes import c_int, create_unicode_buffer, HRESULT, windll
 from ctypes.wintypes import DWORD, HANDLE, HWND, LPWSTR
+
+import eg
 
 GetTempPathW = windll.kernel32.GetTempPathW
 GetTempPathW.restype = DWORD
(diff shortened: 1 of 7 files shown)


kdschlosser:
OK this last commit should finish this up. I am not sure if you changed code climate to bypass the errors. I did a commit on this adding the UTF-8 BOM but if you changed codeclimate to bypass the issue then I am not sure if that was the solution.

topic2k:
Did you see the build error?
```python
  File "C:\projects\eventghost\eg\Core.py", line 354, in TracebackHook
    eg.log.PrintTraceback(excInfo=(tType, tValue, traceback))
AttributeError: 'NoneType' object has no attribute 'log'
Original exception was:
  File "_build\Build.py", line 178, in <module>
    MyBuilder().Start()
  File "C:\projects\eventghost\_build\builder\__init__.py", line 215, in Start
    from Tasks import TASKS
  File "C:\projects\eventghost\_build\builder\Tasks.py", line 125, in <module>
    from builder.BuildDocs import BuildChmDocs, BuildHtmlDocs  # NOQA
  File "C:\projects\eventghost\_build\builder\BuildDocs.py", line 30, in <module>
    import eg
  File "C:\projects\eventghost\eg\__init__.py", line 94, in <module>
    import Core  # NOQA
  File "C:\projects\eventghost\eg\Core.py", line 364, in <module>
    eg.pluginManager = eg.PluginManager()
  File "C:\projects\eventghost\eg\Classes\PluginManager.py", line 28, in __init__
    self.ScanAllPlugins()
  File "C:\projects\eventghost\eg\Classes\PluginManager.py", line 76, in ScanAllPlugins
    for dirname in os.listdir(root):
WindowsError: [Error 3] The system cannot find the path specified: u'C:\\ProgramData\\eventghost\\plugins\\*.*'
```

Don't know why appveyor says success.

kdschlosser:
yeah i see it. I am trying to sort the error. i didn't know appveyor said success

kdschlosser:
I do want to make a comment about the appveyor thing because it drives me batty with the no output thing when it's building EG i have not a clue if the thing is stuck or not.. we need to put at the very least some "."'s or something so we know it's doing something and not just sitting there twiddling it's thumbs

kdschlosser:
Finally that looks like a successful build.

topic2k:
The problem with the appveyor output is, that the powershell script executes python, catching the output until python exits and then print the cached output.

topic2k:
When i start EG it immediatly exits. When run with -debug option, EG hangs i get a log full of exceptions.
[Log.txt](https://github.com/EventGhost/EventGhost/files/1117897/Log.txt)

PR #254, opened 2017-06-30
eg.WinApi: Functions to get information about logged in user

kdschlosser:
I have added User to eg.WinApi which has several functions

The user groups will include direct and indirect memberships

User.Name() -  Gets the current logged in username
User.Domain() -  Gets the domain/computer name of the user that is logged in
User.IsLocalAdmin() - True/False if the user is in the administrators group on the local computer
User.IsDomainAdmin() -  True/False if the logged in user is logged into a domain and if that user is in the Administrators group
User.NameSamCompatible() - returns domain/username of the logged in user
User.Groups() - defaults to the logged in username and the local computer returns a list of all of the groups the user is a member of
User.IsDomainLogin() -  returns if the user is logged into a domain


topic2k:
Can you check what format your user.py has, is it really utf-8? Codeclimate shows an decoding error. The same error we had with NamedPipe.py. I think both are files that you newly created. Maybe wrong format type?

kdschlosser:
# -*- coding: utf-8 -*-

is at the top

topic2k:
That line tells python what the coding should be, but the real encoding used to save the file could be another. I'm just asking because in the last past days you added three files to the repo (NamedPipe.py, User.py and PythonPaths.py). On all three files the same error where raised. Or could it be a problem with codeclimate/radon? But why only on new files?

grafik
grafik
grafik

kdschlosser:
this is happening because codeclimate is checking the file BOM and not the tag at the top. it is happening because i am copy and pasting the header for an existing EG file that i hapopen to have open and since i am working from the eg.WinApi directory you will probably notice all of thise files have that little C for the copywrite and other files simply have a c so either 2 things need to take place we will have to remember to save the file at UTF8 encoded or we will have to change codeclimate to not look at the BOM on the file.

kdschlosser:
I was going to do a check to see if the C would have corrected the problem. but I am unsure if you changed something in codeclimate to bypass the error.

topic2k:
i approved the file in codeclimate (before your new commits). Now i can't see if it changed anything. But i don't bother if the build wan't fail and EG runs.

PR #257, opened 2017-07-07
Fixes restart not working from the file dropdown menu

kdschlosser:
Title says it all.

### DIFF
--- a/eg/Classes/App.py
+++ b/eg/Classes/App.py
@@ -201,12 +201,13 @@ class App(wx.App):
         def Do():
             from eg.WinApi.PipedProcess import RunAs
             args = self.GetArguments()
-            args.append("-restart")
+            args.append("-multiload")
             if self.Exit():
                 RunAs(sys.executable, asAdmin, *args)
                 return True
             else:
                 return False
+
         if threading.currentThread() == eg.mainThread:
             return Do()
         else:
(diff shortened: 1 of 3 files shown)


topic2k:
Your fix with `multiload` raises another problem:
- start EG
- select File --> Restart as administrator
- from commandline: EG.exe -restart

 now you have two runniing instances.

kdschlosser:
good catch. I submitted another commit. it seems to have fixed the problem. I changed how EG is checking for a running instance. I now use the named pipe and only use the appmutex to check that EG currently closed properly when restarting and also if there isn't a running instance. see what happened is because of the -multiload there technically speaking is not a running instance because it has not registered a guid. that is why when i used the -multiload it shortcut the whole process. because it couldn't detect a running copy of EG. That being said tho. I am not sure how the pipe handles multiple connections. we will have to test this a little more.

kdschlosser:
ok I think i have this issue all fixed. it was a timing issue between the processes and threads. what was happening is it would only occur if you had debugging turned on. and this is because of the extra amount of time it took EG to get the messages posted. and because of the exception handling no error would show and it would allow the second instance along it's merry way. which would then start another instance of EG.
so what I have it doing now is it looks for this error and then loops until it can send the message when the pipe isn't busy.

kdschlosser:
this was a real bugger to figure out what was going on with it.

PR #260, opened 2017-07-09
[System] - Code Cleanup - System, Cleans up code

kdschlosser:
Code cleanup/formatting
Moves all internationalism to a single Text class in __init__

### DIFF
--- a/plugins/System/ChangeDisplaySettings.py
+++ b/plugins/System/ChangeDisplaySettings.py
@@ -22,19 +22,9 @@ import wx
 import eg
 from eg.WinApi.Display import GetDisplay, GetDisplays
 
-class ChangeDisplaySettings(eg.ActionBase):
-    name = "Change Display Settings"
-    description = "Changes display settings."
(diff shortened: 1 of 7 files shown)


topic2k:
That's a big PR with lot of changes. I would prefer to have it split, either as more commits or more PR's:
- white space changes
- import changes
- text class changes
- code changes (variable names, line splits, etc.)
- (other?)

kdschlosser:
can do. would you like them as PR's or commits?

kdschlosser:
Oh one other thing. I would actually prefer is each file had it's own text class I would have to go and do some legwork to see if that would be possible to do with breaking the existing language files. but if it does break the the API then how the text classes are handled could get changed in a manner that could handle it. that would be something we could work on in the future and for now this would be fine. but I am making a comment about it for reference.

topic2k:
One PR with serveral commits is ok.

kdschlosser:
Going to redo do to single commit for all changes.

PR #263, opened 2017-07-13
CLI: fix bug when showing message on -netsend failure

topic2k:
Fix bug when showing message on send failure.
Avoid exception and show a message when not enough parameters were given.

### DIFF
--- a/eg/Classes/NetworkSend.py
+++ b/eg/Classes/NetworkSend.py
@@ -23,6 +23,17 @@ from hashlib import md5
 
 ENCODING = locale.getdefaultlocale()[1]
 
+
+def ShowError(msg):
+    app = wx.PyApp()
+    wx.MessageBox(
+        message=msg,
+        caption="EventGhost - warning",
+        style=wx.ICON_EXCLAMATION | wx.OK,
+        parent=None
+    )
+
+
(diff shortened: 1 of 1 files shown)


eg/Classes/NetworkSend.py
```diff
@@ -23,6 +23,17 @@
...
+
+def ShowError(msg):
+    app = wx.PyApp()
+    wx.MessageBox(
+        message=msg,
+        caption="EventGhost - warning",
+        style=wx.ICON_EXCLAMATION | wx.OK,
+        parent=None
+    )
+
```
kdschlosser: only think i would change is in liue of using wx for the dialog use ctypes. the reason for this is because if down the road we make a change we will never end up with conflicting wx.App instances. Now i don't know if that would ever happen. But it seems to be a kind of how things went with EG. anything that produces a dialog before the Core.py file is loaded is done through the use of ctypes

    import ctypes

    ctypes.windll.user32.MessageBoxA(
        0,
        msg,
        "EventGhost - warning",
        0 | 40000
    )

topic2k: That is a good argument. We just need a MessageBox so there is no need to load full wx system. I will change it.

PR #267, opened 2017-07-28
Disable the OK button in configure dialogs on first appearance

kdschlosser:
This will disable the OK button for plugin and action configure dialogs only when it is a new item added to the tree. it will enable the button once a setting change has been made. If the item being configure is one that was already present in the tree then the behavior is the same as it usually is.

This stops the user for inadvertently creating an "empty" or a non configured action

### DIFF
--- a/eg/Classes/ButtonRow.py
+++ b/eg/Classes/ButtonRow.py
@@ -109,8 +109,6 @@ class ButtonRow(object):
         self.numSpecialCtrls += 1
 
     def OnApply(self, event):
-        if hasattr(self, 'okButton'):
-            self.okButton.Enable(True)
         if hasattr(self.parent, "OnApply"):
             self.parent.OnApply(event)
         else:
--- a/eg/Classes/ButtonRow.py
+++ b/eg/Classes/ButtonRow.py
@@ -109,6 +109,8 @@ class ButtonRow(object):
         self.numSpecialCtrls += 1
 
     def OnApply(self, event):
+        if hasattr(self, 'okButton'):
+            self.okButton.Enable(True)
         if hasattr(self.parent, "OnApply"):
             self.parent.OnApply(event)
         else:
--- a/eg/Classes/ConfigPanel.py
+++ b/eg/Classes/ConfigPanel.py
@@ -52,6 +52,9 @@ class ConfigPanel(wx.PyPanel, eg.ControlProviderMixin):
         self.resultCode = None
         self.buttonsEnabled = True
         self.dialog.buttonRow.applyButton.Enable(False)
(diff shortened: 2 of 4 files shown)


GruberMarkus:
I can see this change in version 0.50-rc5, but it is confusing to me: When adding a plugin, the configuration dialog is opened and both, OK and Apply buttons are disabled. Only the Cancel button is active.

If the options presented in the very first run of the configuration dialog are exactly what the user wants, he has to change an option to enable OK and Apply buttons and then revert the unwanted change.

Is this the intended behavior?

topic2k:
Yes, that is the intended behaviour.

It could happen that a plugin has no perfect default values and then fails with an exception to start. Do minimize this happening we force the user to change something before he can just click OK. That way we hope the user think a moment about the values before proceeding. I know, it's not perfect, but we are always open for other (better?) ideas. ;)

kdschlosser:
it was a boo boo. geesh! LOL

funny thing is I have been running a version with this already added and I never had the issue come up. not until the day after it got released. months I have been using EG with this and not stumbled across the issue.

The reason why this cam about is because of actions not having defaulted values. it caused some kind of a batty issue. If I remember correctly it actually had to do with the Jump Action. But now for the life of me I can't remember the exact reason for it. And I know it happened a long time ago. and the issue has not come up since. so maybe this PR simply needs to be reversed. and we can go from there. I still have to look at the restart and see what I broke when I changed the named pipe out for the local host socket connection.

topic2k:
I must say that it annoys me too and i vote for reverting the change. Instead we should 'fix' those config panels that have no usable default values. Either set them to have usable default or enable/disable the buttons itself. Something like:

```python
class MySuperPlugin(eg.PluginClass):

    def Configure(self, data1=''):
        panel = eg.ConfigPanel()
        tctrl = panel.TextCtrl(data1)
        panel.AddCtrl(tctrl)

        panel.EnableButtons(False)

        while panel.Affirmed():
            val1 = tctrl.GetValue()
            
            # validate all settings
            if val1.lower() == 'enable':
                panel.EnableButtons(True)
            
            panel.SetResult(val1)
```

(ConfigPanel already has the EnableButtons function.)

kdschlosser:
This is what I think we need to do. as you will notice I do not think it is really a problem with the OK button being grayed out. it is the annoyance of having to change something in order for it to become active,

we need to enable to apply button. and upon clicking on the apply button the OK button will become available.

what do you think about doing that?

kdschlosser:
I am trying to think of where to go about catching an error if thee apply button is pressed and there is no value that has been set. so the program won't crash.

PR #268, opened 2017-07-30
Use eg.APP_NAME instead of installation folder name for configDir

kdschlosser:
The creation of the config and programdata directories was changed to use the installation folder name. This was an accidental change but it has caused confusion (mainly with me) as to where to look for the config/plugins. there was also an error in it where the debugging log would still be written to the EventGhost config directory. so information was getting written all over the place.


kdschlosser:
I hope you tested this.. you should know by now my code lately seems to be having issues

PR #271, opened 2017-07-31
[EventGhost] - Enhancement - DesktopDisplay, convience module for display attributes

kdschlosser:
DesktopDisplay:
convience module for display attributes

methods:
Refresh() - refreshes the display list. this automatically gets called when EnumDisplays() or GetDisplay() is called.
EnumDisplays() - returns a list of displays
GetDisplay() - you can either pass this the monitor number or monitor name and this will return a DesktopDisplay.Display instance

attributes
displays - this is a list of all of the available displays that are attached
size - this is the virtual size of the desktop display
displayCount - number of connected displays
isTouchScreen - take a guess
isMultiTouch - take a guess
maxTouchCount - maximum number of simultaneous touches if the display is touch capable otherwise 0

DesktopDisplay.Display: 
this class will be returned as a representation of a specific display that is attached.

attributes:
blackout - set True/False to blackout the display
x - upper left corner of the display pixel position relative to the width of the desktop
y - upper left corner of the display pixel position relative to the height of the desktop
w - display pixel resolution width
h - display pixel resolution height
primary - True/False if the display is the primary display
number - display number, this the the number seen in the windows screen resolution dalog
name - nameof the display. this is the name seen in the screen resolution dialog
size - this is the size of the display returned as (w, h)
pos - this is the upper left corner pixel position of the display returned as (x, y)

If for example you have 3 displays attached and you disconnect display #2 this will automatically cause Windows to reorder the displays this will also reorder as required. 
If the screen is blacked out during this reordering it will remove the blackout.
If you change the display position when the screen is blacked out, the blackout should remain

Added 2 actions to the System plugin that will enable the blacking out of a specific monitor in a multi monitor arrangement. this is beneficial in the world of HTPC's because if you have a TV attached at the same time you were not able to shut off the monitor while leaving the TV on. this will not shut off the monitor either but at least it will make it less noticeable.

Changed eg.DisplayChoice  to list the display "number: display name" instead of "Monitor 1", "Monitor 2"
if you use GetStringSelection() it will return the usual "Monitor 1" this is for backwards compatibility with plugins. I have added GetDisplaySelection() to return the actual monitor name. This will not return the display "number: " that is seen in the list it will return only the display name

### DIFF
--- a/eg/Classes/DesktopDisplay.py
+++ b/eg/Classes/DesktopDisplay.py
@@ -1,0 +1,282 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of EventGhost.
+# Copyright © 2005-2016 EventGhost Project <http://www.eventghost.net/>
+#
+# EventGhost is free software: you can redistribute it and/or modify it under
+# the terms of the GNU General Public License as published by the Free
+# Software Foundation, either version 2 of the License, or (at your option)
+# any later version.
+#
+# EventGhost is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+# FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for
+# more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with EventGhost. If not, see <http://www.gnu.org/licenses/>.
+
+
+import win32api
+import wx
+
+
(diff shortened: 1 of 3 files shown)


kdschlosser:
I do not know if you have had the chance to take a peek at this. It is a nice convince class for handling the monitors. we can even pull in changing the display settings if we want to.

PR #276, opened 2017-10-07
Eventghost - Bugfix - Fixes bug in SpinNumCtrl

kdschlosser:
Fixes bug noted here
http://www.eventghost.net/forum/viewtopic.php?f=15&t=9828
and here
http://www.eventghost.net/forum/viewtopic.php?p=48572#p48572


kdschlosser:
no can do boss. I tried doing that. as it states there is a bug in the num control. remember me modifying the spinumctrl some time back and i did the exact same thing. and then users started having problems. so i had to revert it #188

PR #277, opened 2017-10-14
Add Event before the loading of the plugins.

kdschlosser:
Adds  EventGhost.Startup event to take place before the running of the Autostart macro. This is done so that if the user needs to set into place any globals that a script may depend on it can be done before the possibility of a plugin starting and generating an event before the Main.OnInit gets triggered.

There is a payload with this event that is an instance of an AutostartItem.Time() object. I stored a  datetime.datetime object and it returns all methods/attributes correctly except __str__ and __repr__ methods so it would produce a nice date in the log.
example:    EventGhost.Startup Saturday, October 14th, 2017 @ 01:56:46 AM

### DIFF
--- a/eg/Classes/AutostartItem.py
+++ b/eg/Classes/AutostartItem.py
@@ -38,7 +38,7 @@ class Time(object):
             t_fmt = t_fmt.format('th')
         else:
             t_fmt = t_fmt.format(
-                ["st", "nd", "rd"][t.tm_mday % 10 - 1]
+                ["st", "nd", "rd"][self.day % 10 - 1]
             )
         return self.strftime(t_fmt)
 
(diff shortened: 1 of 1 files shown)


eg/Classes/AutostartItem.py
```diff
@@ -20,6 +20,30 @@
...
+    def __getattr__(self, item):
+        return getattr(self._datetime, item)
+
+    def __str__(self):
+        t_fmt = '%A, %B %d{}, %Y @ %H:%M:%S %p'
+        if 4 <= self.day <= 20 or 24 <= self.day <= 30:
+            t_fmt = t_fmt.format('th')
+        else:
+            t_fmt = t_fmt.format(
+                ["st", "nd", "rd"][t.tm_mday % 10 - 1]
```
topic2k: `NameError: global name 't' is not define`

kdschlosser: could be let me look at the code. i could have messed it up somehow. tho i did test it and it worked. maybe i missed the error when it happened.

kdschlosser:
OK sir. I have fixed the traceback issue.

PR #278, opened 2017-10-14
Trigger event Windows.User.USERNAME on EventGhost start

kdschlosser:
Adds the event EventGhost.LoggedInUser.USERNAME on startup
This event is the first event hat will occur. This will allow for macros to be run depending on the username but also if they are administrator or not.

There is a payload that is a dict(IsLocalAdmin=bool, IsDomainLogin=bool, IsDomainAdmin=bool)
IsDomainAdmin will only be added if IsDomainLogin is true
example:    EventGhost.LoggedInUser.Administrator {'IsLocalAdmin': True, 'IsDomainLogin': False}

### DIFF
--- a/eg/Classes/ActionThread.py
+++ b/eg/Classes/ActionThread.py
@@ -20,6 +20,7 @@ from time import clock
 
 # Local imports
 import eg
+from eg.WinApi import User
 
 EVENT_ICON_INDEX = eg.EventItem.icon.index
 
@@ -102,6 +103,20 @@ class ActionThread(eg.ThreadWorker):
                 print "   -", pluginInfo.name
             print "If you want to use them, please add the missing plugins."
 
+        payload = dict(
+            IsLocalAdmin=User.IsLocalAdmin(),
+            IsDomainLogin=User.IsDomainLogin()
+        )
+
+        if payload['IsDomainLogin']:
+            payload['IsDomainAdmin'] = User.IsDomainAdmin()
+
+        event = eg.EventGhostEvent(
+            prefix='EventGhost',
+            suffix='LoggedInUser.' + User.Name(),
+            payload=payload
+        )
+        event.Execute()
         eg.programCounter = (eg.document.autostartMacro, None)
         eg.RunProgram()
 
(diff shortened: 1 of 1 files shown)


eg/Classes/ActionThread.py
```diff
@@ -102,6 +103,20 @@ def StartSession(self, filename):
...
+
+        if payload['IsDomainLogin']:
+            payload['IsDomainAdmin'] = User.IsDomainAdmin()
+
+        event = eg.EventGhostEvent(
+            prefix='EventGhost',
+            suffix='LoggedInUser.' + User.Name(),
+            payload=payload
+        )
+        event.Execute()
```
topic2k: Why are you not directly calling `eg.TriggerEvent`?

kdschlosser: This is a complex issue. But in short I did not want EG to continue loading until after whatever it is that needs to be changed by this event gets done. as an example. In a remote desktop environment some plugins like the Audioendpoint plugin do not function and cause EG to hang. so if you set up a specific username that you will use to login via RDP when EG starts this event gets triggered indicating the username so it would disable the plugin before it loads. and because the autostart macro gets processed differently and doesn't get added to the action thread queue the load of EG needs to stop loading take care of the tasks associated with this event and then continue. at this point if audio endpoint has been disabled then the autostart macro will not try to load the plugin.

kdschlosser:
the username is a convenience thing so you would not have to create a script to find out what username is in the event. see the comment above for use case

kdschlosser:
I do not know why this got closed. maybe an accident???

or if intentional can i get an explanation as to why?

topic2k:
Uups, sorry, didn't realized i closed it.

kdschlosser:
no worries m8. didn't know if you  simply hated the idea that much. LOL

topic2k:
> the username is a convenience thing so you would not have to create a script to find out what username is in the event. see the comment above for use case

I understand and i didn't meant to remove the username, just add a second event where the username is in the payload. That way i can start marcros globaly, regardless of the user(name) that logged in.

kdschlosser:
that would be the purpose of EventGhost.Startup event :P

kdschlosser:
I just wanted to let you know I have been messing around with trying to get EG added into task scheduler. and this may be the method to allow for EG to run without needing to login to the computer. (directly) we will see if i can get it running

topic2k:
I wonder if it makes more sense to change the event prefix to _System_ or _Windows_ instead of _EventGhost_, because the user logged into windows (and not EventGhost)?

kdschlosser:
That's true I can do that.. 

Windows.CurrentUser.USERNAME

PR #281, opened 2017-10-25
[EventGhost] - Enhancement/Bugfix - Replaces the startup shortcut.

kdschlosser:
This replaces the creation of the startup icon with the creation of a task in the Task Scheduler. 
This puts into place a task in the TaskScheduler to start EG with -event OnInitAfterBoot Another Task is will have to be added for starting EG after login. Also need to add the removal of the task when EG is uninstalled. This is a test run to  make sure everything works properly. From the way this looks it should start EG after the computer boots no matter if a user logs in or not. I am not sure.

### DIFF
--- a/eg/Classes/OptionsDialog.py
+++ b/eg/Classes/OptionsDialog.py
@@ -63,7 +63,6 @@ class OptionsDialog(eg.TaskletDialog):
         text = Text
         config = eg.config
         self.useFixedFont = config.useFixedFont
-        task = eg.RegisterTask()
 
         eg.TaskletDialog.__init__(
             self,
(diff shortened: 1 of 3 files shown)


kdschlosser:
I have updated this to allow for complete removal of the startup shortcut for EG as well as deleting the tasks when an uninstall of EG takes place. and several other things as well. see the commit message for details.

kdschlosser:
I am not 100% sure on this. but this may allow for windows to boot and EG to run without the need to make it a service. Being that I am telling the task to login to run EG and also setting the allowed time for the task to run to infinite. I am not sure how windows deals with user logins to run programs from the task scheduler.

kdschlosser:
Not going to use this. Would like to steer away from the use of pywin32

PR #285, opened 2017-10-31
Enhancements to the output in the loglist (colours, icons)

kdschlosser:
I have Split the debugging messages into warning messages and debugging messages. This is because of people turning on the debugging from the drop down and when debugging is turned on tracebacks and messages that occur cause people to submit bug reports when they are not needed.

I also changed the logging colors for debug messages as well as added an icon for it. this provides a better indicator to the user that it is a debugging message and not an error. The warning messages use the same icon as the notice icon but the color of the text is different. so the user is able to again identify who is what.

TriggerEventWait was one of those mechanism that people are false reporting errors
I also changed eg.Utils.AssertInMainThread and eg.Utils.AssertInActionThread and eg.\_\_setattr\_\_ so these will no longer cause EG to stop functioning they now use the warning mechanism. and allow the program to continue along it way.

I set it for EG 0.5 I think this has enough to do with recent changes to warrant it being added to this up coming release


eg/Classes/EventThread.py
```diff
@@ -146,7 +146,7 @@ def Transfer():
         self.AppendAction(Transfer)
         executed.wait(5.0)
         if not executed.isSet():
-            if eg.debugLevel:
-                eg.PrintDebugNotice("timeout TriggerEventWait")
-                traceback.print_stack()
+            eg.PrintWaraningNotice(
```
topic2k: Typo: I think this should be `eg.PrintWarningNotice`

eg/Classes/MainFrame/LogCtrl.py
```diff
@@ -55,25 +73,12 @@ def __init__(self, parent):
...
-        self.attr3.BackgroundColour = oddColour
-        self.attr3.TextColour = (255, 0, 0)
-
-        self.attr4 = wx.ListItemAttr()
-        self.attr4.BackgroundColour = sysColour
-        self.attr4.TextColour = (255, 0, 0)
+        self.attr1, self.attr2 = _create_colour_attributes(sysTextColour)
+        self.attr3, self.attr4 = _create_colour_attributes((255, 0, 0))
+        self.attr5, self.attr6 = _create_colour_attributes((89, 131, 165))
+        self.attr7, self.attr8 = _create_colour_attributes((206, 151, 22))
```
topic2k: Don't kick me 🤓 but i find the colors a bit difficult to read. What about changing the background color for warnings and tracebacks?
```
        # Normal
        self.attr1 = wx.ListItemAttr(colBack=eg.colour.GetOddLogColour())
        self.attr2 = wx.ListItemAttr()
        # Traceback
        self.attr3 = wx.ListItemAttr(colBack=(255, 160, 160))
        self.attr4 = wx.ListItemAttr(colBack=(235, 140, 140))
        # Debug
        self.attr5 = wx.ListItemAttr(
            colText=(50, 40, 190),
            colBack=eg.colour.GetOddLogColour()
        )
        self.attr6 = wx.ListItemAttr(colText=(50, 40, 190))
        # Warning
        self.attr7 = wx.ListItemAttr(colBack=(255, 235, 90))
        self.attr8 = wx.ListItemAttr(colBack=(235, 215, 70))
```
(I've also changed the debug text color a little)

topic2k:
The notice and the warning icons are the same. I vote for another icon for notice: 
info

kdschlosser:
well I vote to take the icon from eg.Print since no one uses that function and come up with some kind of an icon for eg.Print.

OK as far as the colors go. but i am going to have to rewrite the bit that figures out the odd bar color to accept a color being passed into it. I am thinking an orange background with the black lettering for the warnings. and debugging I have not a clue what to use.

kdschlosser:
alright buddy. I changed up the colors. I had to modify Colors.py to do this to allow for input to the GetOdd method. I modified the Print Icon to be a balloon with 3 dots in it and I changed up the Print Notice to be the balloon with the I in it I may change the opening messages to use PrintNotice instead of Print. wanted to get your input on this

topic2k:
I like the icons. The background colour for debug messages i would leave default or use very very light blue (o know you like green 😆 ). The now shiny green is IMO a bit too eye catching, the normal logs getting out of view. For warning and errors/tracebacks it's ok to catch the attraction. That said i favour the (light) red backround/black text for tracebacks.

kdschlosser:
ok you pick the colors. send me the RGB values. and we will roll with those..

and I HATE green I can't stand it. But for some reason it has become EG's unofficial color. this happened before my time and it's not my place to change it. It is almost as difficult to work with as yellow. yellow is almost impossible to see. but green has not a damned color other then black that looks good with it.

kdschlosser:
I know ya like the debugging icon.. I made that myself.. took me a while to do that. it's hard to make an icon so you know what it is.. in only 16x16.

eg/Classes/Log.py
```diff
@@ -38,6 +38,42 @@
...
+    if icon == DEBUG_ICON:
+        msg = 'DEBUG: ' + msg.replace('\n', '\nDEBUG: ')
+
+    elif icon == WARNING_ICON:
+        msg = 'WARNING: ' + msg.replace('\n', '\nWARNING: ')
+
+    std_msg = (
+        strftime("%H:%M:%S: ") +
+        msg.replace('\n', '\n' + strftime("%H:%M:%S: "))
+    )
```
topic2k: I have to add ` + "\n"` here or in pycharm console all output will be on one line. Is it only on machine?

kdschlosser: I am not sure what you mean... It works fine when i run EG

kdschlosser: I got ya. The output into log.txt was all on one line. because i forgot to add the '\n'. I have fixed the problem

kdschlosser:
the bug label was on there because technically speaking the debug output colors should not have been the same as the error colors and this was causing some false bug reporting.. so it has one foot in the bug arena LOL. don't know if you agree on this or not.. I know this has been merged and it really doesn't make a difference. but for correctness in the history you may want to have it there.. all depends on what you think.

PR #286, opened 2017-10-31
[EventGhost] - Enhancement - Utils.EqualizeWidths, 

kdschlosser:
eg.EqualizeWidths was coded to accept a single iterable of controls. I have found this very limiting.

So what I have done is it uses *args and you can pass single controls as positional areuments, or you can pass nested lists/tuples or a combination

These work exactly like before with no changes

    eg.EqualizeWidths(ctrl1, ctrl2, ctrl3, ctrl4)
    eg.EqualizeWidths((ctrl1, ctrl2, ctrl3, ctrl4))

below will set the widths the same for all of the controls in tuple 1. and then it will set the widths the same for tuple 2. tuple 1 will not have the same widths as tuple 2

    eg.EqualizeWidths((ctrl1, ctrl2, ctrl3, ctrl4), (ctrl5, ctrl6, ctrl7, ctrl8))

in the example below the tuples works the same as above but in this example you see ctrl1 and ctrl5 are not in a tuple. those 2 controls will have matching widths but they will not match the widths of tuple 1 or tuple 2

    eg.EqualizeWidths(ctrl1,(ctrl2, ctrl3, ctrl4), ctrl5, (ctrl6, ctrl7, ctrl8))

### DIFF
--- a/eg/Utils.py
+++ b/eg/Utils.py
@@ -258,10 +258,23 @@ def EnsureVisible(window):
     # set the new position and size
     window.SetRect((left, top, right - left, bottom - top))
 
-def EqualizeWidths(ctrls):
-    maxWidth = max((ctrl.GetBestSize()[0] for ctrl in ctrls))
-    for ctrl in ctrls:
-        ctrl.SetMinSize((maxWidth, -1))
+def EqualizeWidths(*args):
+
+    def iter_ctrls(ctrls):
+        maxWidth = -99
+        for ctrl in ctrls:
+            if isinstance(ctrl, (list, tuple)):
+                iter_ctrls(ctrl)
+                continue
+            else:
+                maxWidth = max((ctrl.GetBestSize()[0], maxWidth))
+
+        if maxWidth != -99:
+            for ctrl in ctrls:
+                if not isinstance(ctrl, (list, tuple)):
+                    ctrl.SetMinSize((maxWidth, -1))
+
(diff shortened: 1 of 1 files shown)


topic2k (approved):
LGTM
I think we can include it 0.5 as it is only a small change.

kdschlosser:
I am going to run it through a full gambit of tests to make sure everything is OK with it.

kdschlosser:
OK I gave it a good test. It's good to go.

kdschlosser:
do you understand what the benefit to this would be?

You have the ability to create a horizontal sizer add the label and then the control. and you will only have to make the call to EqualizeWidths one time. bundle all the label controls together and bundle all of the other controls together and pass then at the same time.

If you use eg.BoxedGroup it becomes really easy. because you do not need to create a variable for each label. you can pass the instance of the label directly to BoxedGroup then do 

     eg.EqualizeWidths(box.GetColumnItems(0), box.GetColumnItems(1))

and for the real trip. 

    while panel.Affirmed():
        panel.SetResult(*(item.GetValue() for item in box.GetColumnItems(1)))

This would work if all of your controls have a GetValue method
if all of the controls are in the ControlProviderMixin then i believe they all have a GetValue

So now you can pass the labels and the controls directly to the BoxedGroup without the need to create variables for them. This would be used for very simple dialogs but it is a really nice convenience.

PR #289, opened 2017-12-17
Fix -event CLI option

topic2k:
The -event option were unintentional changed. The command `EventGhost.exe -e myevent with multiple payloads` generates the event `multiple.myevent` with payload `u"with"`. The correct (as in EG 0.4.r1722) event should be `Main.myevent` with payload `[u'with', u'multiple', u'payloads']`.


topic2k:
With
```EventGhost.exe -e MYEvent payload1 payload2 "this is payload3" -debug```
 you get payload
 `['payload1', 'payload2', 'this is payload3']`

PR #291, opened 2017-12-19
[EventGhost] - Enhancement - NamedPipe.py, Removes sending a ping through the pipe

kdschlosser:
I have removed the need to ping the named pipe to check the existence of an already running session of EG. This new method allows for faster loading times of EG. It is also less error prone. Another added benefit is PR #290 runs about a half a second faster when triggering 100 events from the command line.


kdschlosser:
if ya had merged this one first it would have merged ok. I think. no worries tho i will correct it.

kdschlosser:
OK I have fixed the issues with this but it still has that blasted message up about the conflicts and also appveyor fails without even running at all.

PR #296, opened 2018-02-06
[requests] PythonPaths.py, Fixes SSL error

kdschlosser:
Adds the environment variable REQUESTS_CA_BUNDLE which points to the requests/cacert.pem file. This should take care of an SSL error that comes up on some systems.

http://www.eventghost.net/forum/viewtopic.php?f=15&t=10066&p=50080#p50080

http://www.eventghost.net/forum/viewtopic.php?f=2&t=9725&start=15&sid=600e97c0d6743203ab78f4316b35917f

we should wait on merging this until it gets confirmed as a working solution.


kdschlosser:
This is a working solution. YAY!

PR #303, opened 2018-03-01
Add exception catching to plugin close.

kdschlosser:
This is so that EG doesn't get hung up forcing a user to terminate the
process potentially leading to things like orphaned ports. or files not
written to when EG closes (config.py)

### DIFF
--- a/eg/Classes/PluginInstanceInfo.py
+++ b/eg/Classes/PluginInstanceInfo.py
@@ -51,7 +51,11 @@ class PluginInstanceInfo(PluginModuleInfo):
         if self.isStarted:
             self.Stop()
         if not self.initFailed:
-            self.instance.__close__()
+            try:
+                self.instance.__close__()
+            except:
+                import traceback
+                eg.PrintDebugNotice(traceback.format_exc())
 
     def CreateInstance(self, args, evalName, treeItem):
         self.args = args


eg/Classes/PluginInstanceInfo.py
```diff
@@ -51,7 +51,11 @@ def Close(self):
         if self.isStarted:
             self.Stop()
         if not self.initFailed:
-            self.instance.__close__()
+            try:
+                self.instance.__close__()
+            except:
+                import traceback
+                eg.PrintDebugNotice(traceback.format_exc())
```
topic2k: I would prefer `eg.PrintTraceback()` because this will be shown in log even if you don't run with -debug enabled.
(I think it's better to show the error instead of hiding it. If no one sees the error, no one can try to fix it.)

kdschlosser: I was thinking about this. It can get stuck. also I do not remember if this gets called from the action thread or not. Need to check that. because of how sensitive windows has become with registering for notifications. I think it to be wise to also have the same thread run the close routine. just to be safe. what we need to do to to use the call wait and set a default timeout. this way if something is wrong it will not cause EG to get all kinds of stuck. Since there is no mechanism to "end a thread" I would like to write a wrapper for the threading class that would set all threads made in EG as daemon threads. we would have to remove the line where EG skips waiting for those threads to close. this way a standard close routine will work properly. but in the event it does not the thread will still terminate. 

The reason i didn't use PrintTraceback was because of the mangling it does. I have never really gone through it to see what exactly gets changed. but it is not a standard traceback output.

If i could get my blasted github to work properly i would push up some changes i have made. damned thing has some kind of a server certificate issue. i believe it is related to TLS 1.2 but everything I have tried has been a fail

topic2k: So we have two choices:

- Set the milestone for this PR to 0.6 and wait till you've done that wrapper thing.
- Merge the PR as it is, so there is at least a little exception handling until 0.6.

I would go with the latter.

kdschlosser: I do agree. on that. I still have not gotten my github to work properly. I do not know what it's issue is.  so if you wanted to change it to eg.PrintTraceback you are going to have to make the change. I am game for having the information printed out in the log as well. so whatever you would like to do.

PR #309, opened 2018-04-15
BuildChangelog.py: fix UnicodeEncodeError

topic2k:


kdschlosser:
you are going to need to do a little more work on the unicode thing. I made up this code block. It handles unicode as well as python 2 or 3

            try:
                file_data= data.decode('utf-8')
            except UnicodeEncodeError:
                try:
                    file_data = data.decode('latin-1').decode('utf-8')
                except UnicodeEncodeError:
                    file_data = data
            except AttributeError:
                file_data = data

So you will want to decode the changelog to determine how it should be written. Took me a while to get this masterpiece right LOL. This is because utf-8 doesn't work for all unicode types. but when coupled wit the latin-1 it does. and in python 3 it is all handled internally so no worries there.

Issue #232, opened 2017-05-21
Access denied on updating plugins via import of egplugin file


ThomasBott:
I'm getting an access denied error when I try to update one of my plugins that's already installed by using an egplugin file.

23:56:36   Traceback (most recent call last) (0.5.0-rc2):
23:56:36     File "wx\_core.pyc", line 16766, in <lambda>
23:56:36     File "C:\Program Files (x86)\EventGhost\eg\Utils.py", line 441, in LogItWithReturnWrapper
23:56:36       result = func(*args, **kwargs)
23:56:36     File "C:\Program Files (x86)\EventGhost\eg\Classes\PluginInstall.py", line 169, in Import
23:56:36       shutil.rmtree(info.path, False)
23:56:36     File "shutil.pyc", line 247, in rmtree
23:56:36     File "shutil.pyc", line 252, in rmtree
23:56:36     File "shutil.pyc", line 250, in rmtree
23:56:36   WindowsError: [Error 5] Zugriff verweigert: u'C:\\ProgramData\\EventGhost\\plugins\\AudioEndpoint\\AudioEndpointControl\\EndpointvolumeAPI.py'

Maybe it's because I import a library along with the plugin, didn't check that in detail.

Import works then I remove the plugin folder before importing.

EG 0.5 RC 2

kdschlosser:
this might be because you are using windows 10 and have the plugin installed into the core plugins directory. and when you go to install via the egplugin file it installs the new/updated version into the user plugins directory. and then will try to delete the old directory that is in the core plugins directory and windows will not allow an application to modify the program files directory

ThomasBott:
Nope, just double checked it, there is none in ProgramFiles, only one in ProgramData...
And it works when I delete that one manually

kdschlosser:
hmmm.... if the PR for all of  the work I have done to the plugin install had been merged we could have had a more detailed error message
but i am really leaning twords a permissions thing. as an example if you have installed a plugin whether manually or by using the eg plugin file as an administrator. but are then trying to update the plugin as a normal user you could get an error upon removing the plugin. this wuld be due to permissions. 

Are you running windows 10? 
Do you think that possibly the above scenario could have taken place?

[...]

PR #310, opened 2018-04-22
[EventGhost] Bugfix/Enhancement NamedPipe.py Rewrite

kdschlosser:
Complete rewrite. This is now a multi-threaded asynchronous overlapped IO pipe written in pure python using only the std lib. I have added commented lines detailing how this whole system works.


kdschlosser:
No longer needed. 
#232 succeeds this PR

PR #313, opened 2018-05-25
[EventGhost] - BugFix - NamedPipe.py, Rewrite in pure python

kdschlosser:
Removes use of pywin32 for handling the named pipe.

Only uses python std lib

Changes handling of commands so only commands directed to eg.document or eg.mainFrame will be run in the main thread.

Changes the closing of the Pipe so the client has to initiate the close by sending CLOSE. this allows the pipe connection to remain open for further commands to be sent over the pipe without the need to create a new connection.

Changes the pipe so a pipe name can be supplied. This is so that the named pipe code is not dedicated to only the use of EG. it can now be used in plugins or a script if needed.

Changes the pipe instance checker. I have created a separate function is_pipe_running that allows the user to supply a pipe name to check for the instance of a named pipe.

Changes the exception creation in the command processor to no longer throw exceptions. Instead if an error occurs the returned value will be one of the following.

CommandFormatError
ParameterFormatError
CommandMalformedError
CommandNotFoundError

There is also the following 2 errors that can be returned. These are only returned in the event of the command that was executed not returning within 5 seconds. because we run the command in 2 different threads the main thread or some other thread the returned error values will reflect that.
MainThreadHung
UnableToProcessCommand

### DIFF
--- a/eg/Cli.py
+++ b/eg/Cli.py
@@ -158,7 +158,7 @@ if args.isMain:
         args.isMain # and
         # not args.pluginFile
     ):
-        if NamedPipe.is_eg_running:
+        if NamedPipe.is_eg_running():
             if args.restart:
                 restart()
             else:
--- a/eg/Core.py
+++ b/eg/Core.py
@@ -62,9 +62,6 @@ eg.CORE_PLUGIN_GUIDS = (
 
 eg.namedPipe = NamedPipe.Server()
 
-if eg.Cli.args.isMain:
-    eg.namedPipe.start()
-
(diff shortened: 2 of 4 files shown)


topic2k:
I wanted to test the WIP, but it didn't start. In CLI.py, line 161, the parenthesis are missing:
`if NamedPipe.is_eg_running():`

kdschlosser:
mannnn i thought i pushed that.

kdschlosser:
ok sir, it is fixed

kdschlosser:
No longer needed. 
#323 succeeds this PR

PR #314, opened 2018-06-07
[EventGhost] - Enhancement - WinApi.SendKeys,  Improves performance in SendRawCodes1

kdschlosser:
The old mechanics of this method would send one key at a time and then wait for the notification to get processed by the receiving window. The user32.SendInput function allows you to send multiple keystrokes at a time. 

I have modified the code so it now sends all of the keystrokes in a single notification and then waits only once. The difference in performance is astonishing.

The xml below is a way of testing this. It is a very strange way you have to go about getting the Send Keystrokes action to use this method for sending keystrokes. You need to use Find Window to locate the window. then you have to bring the window to the front (this gives it focus) then in a python script you need to delete the contents of eg.lastFoundWindow, then run the Send Keys action. This will send the keys using this method. 

I suggest using the code below on an unaltered version of EG so you can see the speed. the code below requires you to have notepad open as this is what it is going to send the keys to. The run the same test using this modified version.

    <?xml version="1.0" encoding="UTF-8" ?>
    <EventGhost Version=" ">
        <Macro Name="Find Window: notepad.exe" Expanded="True">
            <Action>
                Window.FindWindow(u'notepad.exe', None, None, None, u'Edit', 1, False, 0.0, 0)
            </Action>
            <Action>
                Window.BringToFront()
            </Action>
            <Action>
                EventGhost.PythonScript(u'del eg.lastFoundWindows[:]')
            </Action>
            <Action>
                Window.SendKeys(u'This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test', False, 2)
            </Action>
        </Macro>
    </EventGhost>

### DIFF
--- a/eg/WinApi/SendKeys.py
+++ b/eg/WinApi/SendKeys.py
@@ -435,8 +435,20 @@ class SendKeysParser:
         Uses PostMessage and SetKeyboardState to emulate the the virtual
         keycode. Can send to a specified window handle.
         """
+
         keyboardStateBuffer = self.keyboardStateBuffer
-        for block in keyData:
+        for i, block in enumerate(keyData):
+
+            def wait():
+                if len(block) > 1:
+                    self.WaitForInputProcessed()
+                else:
+                    try:
+                        if len(keyData[i + 1]) > 1:
+                            self.WaitForInputProcessed()
+                    except IndexError:
+                        pass
+
             if mode == 1 or mode == 2:
                 for virtualKey in block:
                     keyCode = virtualKey & 0xFF
@@ -460,7 +472,9 @@ class SendKeysParser:
 
                     SetKeyboardState(byref(keyboardStateBuffer))
                     PostMessage(hwnd, mesg, keyCode, lparam)
-                    self.WaitForInputProcessed()
+
+            if mode == 2:
+                wait()
 
             if mode == 0 or mode == 2:
                 for virtualKey in reversed(block):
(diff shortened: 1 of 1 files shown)


kdschlosser:
Ok I have added an auto detection of the window handle that is in focus vs the one that is passed into SendKeys. if they are the same then it will use the better mechanism to send the keystrokes. the way the original code was the better mechanism would only be used if the window handle passed into SendKeys was None.

I have created a new test macro. You can see both ways in action. one you will run when notepad is visible but do not make it active. the second is after you run the macro you will need to make notepad active. specifically the text entry area (the white box) simply clicking on the title bar or the button to make it visible will not active the faster mechanism.

    <?xml version="1.0" encoding="UTF-8" ?>
    <EventGhost Version=" ">
        <Macro Name="Find Window: notepad.exe" Expanded="True">
            <Action>
                Window.FindWindow(u'notepad.exe', None, None, None, u'Edit', 1, False, 0.0, 0)
            </Action>
            <Action>
                EventGhost.Wait(2.0)
            </Action>
            <Action>
                Window.SendKeys(u'This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test.This is a test{Enter}', False, 2)
            </Action>
        </Macro>
    </EventGhost>

This fixes a possible issue of when sending a long string to an active window and you used FindWindow to make sure it was running before sending if the user started typing at the same time the string was being send the 2 would get all mixed up. The fixes that bug. SendInput cannot get mixed up with any other input whether it is from a Windows API call or from the user. 

If keystrokes are being sent to a window that is not active it cannot get jumbled up by user input because it is not the active window. and because of how EG works you are not ale to send multiple message at the same time. So no worries there either.

PR #315, opened 2018-06-07
BuildLibrary.py: fix SyntaxError in py2exe for Jinja2

kdschlosser:
This is a fix for jinja2 > 2.8 SyntaxError when p2exe is byte compiling the files. The traceback occurs because newer versions of jinja 2 have files that include asyncio code.

The way i wrote this fix is it will check the source for any reference to an import of asyncio and if founf it will check the python version. if asyncio is not supported for that version it will return an empty byte compiled file

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -90,11 +90,7 @@ install:
         pip install -q -U "setuptools==34.3.0"
         "    Done."
         PipInstall "wheel 0.29.0" "wheel==0.29.0"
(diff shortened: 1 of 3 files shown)


topic2k:
So this will remove the need for installing jinja2 2.8.1 i think you should also remove it from the appveyor.yml.

kdschlosser:
Yes that is correct. I wanted to make sure that it all checks out OK before making any changes to appveyor. But now that I think about it to show it works properly it should be removed from appveyor

PR #316, opened 2018-06-07
Makes EG more IDE compatible

kdschlosser:
You're gonna love this one, I know I do!

### DIFF
--- a/eg/__init__.py
+++ b/eg/__init__.py
@@ -97,11 +97,133 @@ eg = DynamicModule()
 
 # This is only here to make pylint happy. It is never really imported
 if "pylint" in sys.modules:
+    def RaiseAssignments():
+        pass
+
     from Init import ImportAll
     ImportAll()
+
     from StaticImports import *  # NOQA
+    import Core
+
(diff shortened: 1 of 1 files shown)


kdschlosser:
UMMM... It allows for a more complete integration into an IDE.

OK so right now if you use pycharm and create a project for EG if you then go into any of the file and type in eg.PrintDebudNotice. it shows an error. and brings up none of the arguments. the project is riddled with errors about not being able to locate attributes or variables the class structure layout is incomplete or wrong. 

type in eg.config. and you should get a list of what methods/properties/variables are available for the eg.Config.Config class without this modification you do not. with it you do

this corrects about 95% of those types of issues. Remember how we hate the DynamicModule because it doesn't allow an IDE to work like an IDE?? well this is the bridge that allows the IDE to cross over the gap to get the needed information.

as i said. Best modification ever. without the code autocomplete for function names and argument names i find myself spending way to much time having to go and open the file where that code is located to see what the name is or what the arguments are. when i shouldn't have to do that.

PR #317, opened 2018-06-07
Remove obsolete COMServer

kdschlosser:


kdschlosser:
no. it does not matter
because there is already a named pipe in place it can be done at any time.

PR #323, opened 2018-10-22
Swaps out Named Pipe for Socket

kdschlosser:
I swapped out the problematic named pipe for a socket variant. 

The socket version was a bit tricky because I wanted to make sure it was secure. I did this 3 fold.

The first thing I did was i made it run on the internal loopback of the TCP/IP stack. this will ensure that only something on the local computer can access it. 

The second thing I did was the data sent across the connection is encrypted using AES.

The Third thing was a bit tricky. What I have set up is the generation of the key used is not something anyone is going to have easy access to. and the key gets changed each and every single build. the key is inserted into the CryptoSocket file and that file is then compiled into a dll using Cython. All traces of the source file containing the key are deleted immediately. and the compiled dll is then added to the compile lib folder.

here is an issue with Stackles and Cython. I have worked this problem out and have a specially compiled version of the Stackless Python27.dll file. This file is included in the build/data/python27 folder.
Because cythonizing the CryptoSocket code causes an error if the builder is run using the stock Stackless dll I have the builder do some checks to see if the dll is located in the python installations folder. if it is the file gets renamed. and then the path to the new dll is added to the environment. and the build gts restarted as a subprocess. not before setting a marker in the environment so we do not end up in a loop. after the build is finished. successful or not the original dll gets put back into place. If the dll is not found in the installation folder. then the pah to the new one is added and away we go.


kdschlosser:
Oh yeahhhh... No red X's

@topic2k 
Take a look at this when you get the time. This or a working version of the named pipe needs to be added to the next release of EG. I can't seem to get a handle on why there are issues with the named  pipe.  I am not able to replicate the problems, I do not doubt there is an issue. where it is is beyond me. because it is the same exact issue that occurs in all of the variations I have written. using pywin32 or not. I wonder if there is some ind of a goof in Python or possibly between Python and the windows API. 

I am simply tired of rewriting the thing over and over again. So this is the replacement I have come up with. I do like this solution a lot better. it is far more secure.

topic2k:
I'm going to merge this PR next weekend if i don't hear any _NO_ :smiley: and eventually build a new RC.

kdschlosser:
I have not heard of an issue with it. but please do me a favor and run some tests on it to test it. I slammed it with 200 concurrent calls and it did not error and every event was produced.

I also hit it with several thousand consecutive events and no errors and every event went through. with an a impressive 17ms spacing between events. and some of that is due to the length of time it takes to create the thread and call subprocess on the sending side of things.

But any who give it a good once over and see if you get get it to trip up. Tho I still am not able to get the current release to mess up. so who knows.

topic2k:
I tested it with this command (testesd on Win10 (1803) 64Bit, WinXP SP3 and WinVista Business 32Bit):
`for /l %i in (1,1,100) do eventghost.exe -e evt%i`

This work without problems if EG is already running, when not, there will be a lot instances be started. But because of it's nature i think EG would normally already be running, so i can live with this _bug_.

kdschlosser:
OoOO now I have to think how to go about handling that problem. The only way I can think of would either be a lock file or check the process list. What is happening is you have all of those instances trying to register the mutex.

Issue #319, opened 2018-09-18
Right to Left bug


Romank1:
Hi,

I use Windows 10 v1803 x64 English and  EventGhost v0.5.0 rc4.

When I change "Region Format" to Hebrew:
![snap1](https://user-images.githubusercontent.com/4432037/45708695-c3d1fd00-bb8a-11e8-975c-db5746381ea7.jpg)

I get issue with text location:
![snap3](https://user-images.githubusercontent.com/4432037/45708697-c3d1fd00-bb8a-11e8-9681-7c0fa6bbc92c.jpg)

When I change "Region Format" to English, I don't get this issue.
EventGhost 0.4.1 don't have this issue.

kdschlosser:
I need to have a few questions answered.
does Hebrew read right to left or left to right?
do you have the language in Windows set to Hebrew?
what do you have the Location in Windows set to?

I thought that Hebrew is  written and read from right to left. so that by definition should change the orientation of the text to read from right to left having right justification. I am thinking that this would actually be a bug in 0.4.1 and not in 0.5.0.

I am trying to get a better understanding of your setup. There are 3 areas that effect how windows and it's application will display text. the Format tab the Location tab and the Keyboard/Language tab in the Region control panel. I would need to know what all of these are set to in order to be able to give you a solution/answer

kdschlosser:
I didn't want to leave you ganging with this.

Upon investigation... it's half a BUG..
The text being right justified is not.. open internet explorer and see what happens.

Now the whole frame getting reversed.. like the close button ending up on the left.. That is a BUG.. I posted a bug report on the wxPython repository.. I wouldn't hold my breath on getting it fixed any time soon. They seem to not do anything about bugs unless it completely breaks something. If it is how it is getting displayed that is wrong. they will be in no big hurry to fix it. 

The issue with wxPython is that over the years and having so many fingers in the cookie jar so to speak the code is a complete mess. It is not something I am going to try and fix.. I would end up wanting to re wright the whole thing.

I am going to see if there is a way I can turn it off.. If i can do that then I can set an option in EG for it to ignore the those settings in windows.

kdschlosser:
[...]

PR #325, opened 2018-10-23
Fixes getting the default user language from the OS

kdschlosser:
default language revisited

This is the second time doing this. the first time I could not remember
what the problem was exactly. It is something that is not easily
replicated.

Details:

the description of 

    wx.Locale.GetSystemLanguage

states "tries to detect users default locale setting."
This description is inaccurate. It does not try to get the "users"
default locale. it does exactly what the method name implies.
it gets the "system" default language. which may not be the language
the currently logged in user has set to be displayed o their screen.

The system default language is determined by what was purchased. So if
i live in Russia. the Windows available for purchase there would be a
Russian edition. Now as the user I can tell windows to display whatever
language I want. so to me everything is pretty seamless. If we
installed EG on a Russian copy of Windows where the user has all of the
language settings set to English, EG would display the Russian language.
Not Ideal. We want to display the language that the user is
currently displaying in Windows.

updates option dialog to display language selection in native tongue.


kdschlosser:
The last PR fixes #319

eg/Classes/Translation.py
```diff
@@ -16,145 +16,1015 @@
...
+        'br_BR': u'Breton',
+        'bs_BS': u'Bosanski',
+        'ca_CA': u'Català',
+        'co_CO': u'Corsican',
+        'cs_CS': u'čeština',
+        'cy_CY': u'Cymraeg',
+        'da_DA': u'dansk',
+        'de_AT': u'Deutsch - Österreich',
+        'de_CH': u'Deutsch - Schweiz',
+        'de_DE': u'Deutsches Deutschland',
```
topic2k: It's just
```suggestion
        'de_DE': u'Deutsch',
```

kdschlosser:
I did think of this. But if you follow the rest of the German layouts like 
German - Austrian
German - Switzerland

the German - Germany would be correct no? I am not sure if the syntax is correct. i do remember seeing in the translator that it was a confirmed syntax. But you would know better then I, seeing how you live there and all ¯\_(ツ)_/¯

topic2k:
Ok, if you want to follow the syntax then "Deutsch - Deutschland" would be the correct spelling. I'm ok with either variant.

topic2k:
What's your opinion? Will you change `Deutsches Deutschland` to one of 
- `Deutschland`
- `Deutsch - Deutschland`
- `Deutsch`

kdschlosser:
Which ever way you want. I am not able to make the change. Or I should say I don't want to. I am on my cellphone it will be a pain.  Can you make the change?

topic2k:
Ok, will do.

PR #326, opened 2018-10-23
Fix links to the EventGhost website in Readme

debanjum:
Replace eventghost.org with eventghost.net


kdschlosser:
awesome. forgot about those links.

PR #293, opened 2018-01-03
[Eventghost] - Requirement - Updates copyright notice

kdschlosser:
The PR title says it all.

PR #327, opened 2018-10-27
Update some missed links to eventghost.net

dequi:
Updates urls to eventghost.net that are not covered in PR #293 or PR #326

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,6 +1,6 @@
 # I need help with EventGhost
 
-Then you're in the wrong place. Support requests should be posted on [our forum](http://www.eventghost.org/forum/), either in [General Support](http://www.eventghost.org/forum/viewforum.php?f=2), or, if you need support for a specific plugin, in that plugin's thread in [Plugin Support](http://www.eventghost.org/forum/viewforum.php?f=9). Be sure to search before posting, as the most timely answer to your question will always be the one that's already there.
+Then you're in the wrong place. Support requests should be posted on [our forum](http://www.eventghost.net/forum/), either in [General Support](http://www.eventghost.net/forum/viewforum.php?f=2), or, if you need support for a specific plugin, in that plugin's thread in [Plugin Support](http://www.eventghost.net/forum/viewforum.php?f=9). Be sure to search before posting, as the most timely answer to your question will always be the one that's already there.
 
 
 # I want to report a bug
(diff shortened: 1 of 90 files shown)


kdschlosser:
hey D any way you can do me a favor???

on this specifically the AboutDialog I have yet to go through the rest or even the ones I did up.

any links to the forum change them to 

    http://forum.eventghost.net


and the links to media wiki would be 

    http://wiki.eventghost.net

and any links to the github repository

    http://github.eventghost.net

and the main page

    http://eventghost.net

scratch the www. it is an assumed thing. if no site can be contacted at domain.ext almost all browsers will now add the www automatically. just like the . on the end of every single web address that no one really knows about, it is automatically added.  

i am going to add a 

    http://help.eventghost.net

that points to the documentation.

dequi:
To www or not to www, that's called the **dubayu dubayu dubayu holy war**. Google and Facebook do www, Twitter doesn't. The only issue I could find is regarding cookies on subdomains, others are just preferences. I tend to follow Google in such matters but since you did the copyright links without, I'll adjust them. Current server setup should support them both anyways without the need of the browser trying to guess the right one.

Regarding the _wiki_, _forum_  and _help_ sub-domains: what is the added benefit? You are just introducing an unnecessary redirect. **http://eventghost.net/forum/** is just as long as **http://forum.eventghost.net/** and it requires more dns configuration to support, expand or reconfigure. You will have to convince me on that one.

kdschlosser:
actually there is no DNS settings for setting that up additional hosts. It is just a couple of line in a apache config file nothing to maintain or ever worry about. nothing overly overly complicated. I know it's not any more or less typing. and it is only a matter of a / vs a . I do not know if you agree on this or not,  I do think the use of the hosts look a little bit more professional. looks like there are more servers. 

whichever way you want to go. 

and yes the server is set up for www and not www so no multiple DNS queries needed there. and the www is set up the same way as the other hosts. 

Because we have a VPS we have control of the server. right down to installing the operating system. we install any application we want. Its not a shared hosting plan. so we have dedicated IP addresses. The price per month was above and beyond awesome for the resources we were given. and it removed any and all limitations in a so call unlimited plan.

dequi:
Hi Kevin, are you really sure you want to drop the www? Google top results for EventGhost are all www urls. If so, you might need to setup a 301 redirect from www to non-www or add the canonical URL meta information to the www version to avoid duplicate content SEO punishment. (I would choose the first just like Google, Twitter, Facebook do) For the same reason I would not opt to change to use those subdomains. Also the user won't notice the subdomain urls as it is a link in the code and when clicked the browser will redirect almost immediately and show the http://eventghost.net/forum/ url anyway.

kdschlosser:
whatever you want to do.

dequi:
Just to clear then what I need to do. We stick to the http://eventghost.net/forum type url's without the www and I need to update the PR.  (and you set up a 301 from www to non-www so that Google eventually shows the non www urls in the results). We don't do the different subdomains.

Or

You update the copyright notice url to include the www?

kdschlosser:
i will update the copyright notice to fix the url's

kdschlosser:
Yes sir

PR #330, opened 2018-11-13
[PowerDVD] Update to work with newer PowerDVD versions (v12+)

GruberMarkus:
The exising plugin does not work with newer PowerDVD versions (at least from version 12 upwards).

This update makes PowerDVD 18 controllable from within EventGhost. It is probably also compatible with earlier versions of PowerDVD.

### DIFF
--- a/plugins/PowerDVD/__init__.py
+++ b/plugins/PowerDVD/__init__.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is a plugin for EventGhost.
-# Copyright © 2005-2016 EventGhost Project <http://www.eventghost.org/>
+# Copyright © 2005-2018 EventGhost Project <http://www.eventghost.net/>
 #
 # EventGhost is free software: you can redistribute it and/or modify it under
 # the terms of the GNU General Public License as published by the Free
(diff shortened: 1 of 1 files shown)


plugins/PowerDVD/__init__.py
```diff
@@ -20,58 +20,155 @@
...
+        group.AddActionsFromList(PictureActions, ActionPrototype)
+
+        group = self.AddGroup("Music", "")
+        group.AddActionsFromList(MusicActions, ActionPrototype)
+
+        group = self.AddGroup("General Navigation", "")
+        group.AddActionsFromList(NavigationActions, ActionPrototype)
 
+        group = self.AddGroup("General Program Management", "")
+        group.AddActionsFromList(ProgramManagementActions, ActionPrototype)
```
topic2k: Here is a lot of repeating code that could be rewritten. Put the action lists into a dict:
```python
action_list = {
    'AudioActions': [
        ("Pause", "Pause", "Pauses playback.", "{Space}"),
        ("VolumeUp", "Volume Up", "Increase audio volume.", "+"),
        ...
    ],
    'VideoPlayActions': [
        ("ToggleMute", "Toggle mute", "Mute volume.", "q"),
        ("TogglePlayPause", "Toggle Play/Pause", "Play/pause media playback.", "{Space}"),
        ...
    ],
    ...
}

...

    def __init__(self):
        for grp, actions in action_list.iteritems():
            group = self.AddGroup(grp, "")
            group.AddActionsFromList(actions, ActionPrototype)


```

plugins/PowerDVD/__init__.py
```diff
@@ -20,58 +20,155 @@
...
+        # AHK code for toggling play/pause:
+        #     ControlGet, OutputVar, Hwnd,,, PowerDVD
+        #     ControlSend, , {space}, ahk_id %OutputVar%
+        #
+
+	gWindowMatcherMainWindow = eg.WindowMatcher('PowerDVD{*}.exe', 'PowerDVD', None, None, None, 1, "false", 0, None)
+	gWindowMatcherPlayerWindow = eg.WindowMatcher('PowerDVD{*}.exe', 'PowerDVD', '{*}', '{*}', '{*}', 1, "true", 0, None)
+
+        hwnds = gWindowMatcherMainWindow() + gWindowMatcherPlayerWindow()
+
```
topic2k: `gWindowMatcherMainWindow` and `gWindowMatcherPlayerWindow` needs only to be instantiated once. So you can move them to module level (just like in v1.1)

GruberMarkus:
@topic2k: Thanks for your fast review and valuable input! I have updated the code.

plugins/PowerDVD/__init__.py
kdschlosser: This would be the ideal way to go about adding actions via a list. EG has been coded to accept this format.


    
    
    
    # SendAction is the class we want to use when sending the command. It sets
    # the data to be used in self.value, the name of the class also gets set to what
    # is specified in the list
    
    class SendAction(eg.ActionBase):
    
        def __call__(self):
            # self.value gets put in place when the instance is constructed
            command = self.value
            # we then pass the command to the plugin to be processed
            self.plugin.send(command)
    
    
    # because this list is unchanging it is a constant, and should have the proper
    # naming convention used. which would be all uppercase with an
    # _ representing a space
    
    
    # the layout is as follows
    
    # ACTION_LIST = (
    #     (
    #         group glass always eg.ActionGroup,
    #         group class name,
    #         display name,
    #         description,
    #         tuple of actions to be added (
    #             (
    #                 class to be used for the action,
    #                 name of the class,
    #                 display name,
    #                 description,
    #                 data to be set into self.value
    #             ),
    #             (
    #                 class to be used for next action the same class can be used more then one time.,
    #                 name of the 2nd class has to be unique,
    #                 display name,
    #                 description,
    #                 data to be set into self.value
    #             ),
    #         ),
    #     ),
    #     (add next group and action here)
    # )
    
    
    # eg.ActionGroup is the class that is used for a folder item on the tree
    # we pass that in the list and the core code known what to do from there
    
    ACTION_LIST = (
        (
            eg.ActionGroup,
            'Audio',
            'Audio Controls',
            'Audio Controls',
            (
                (
                    SendAction,
                    "fnPause",
                    "Pause",
                    "Pauses playback.",
                    "{Space}"
                ),
                (
                    SendAction,
                    "fnVolumeUp",
                    "Volume Up",
                    "Increase audio volume.",
                    "+"
                ),
            )
        ),
        (
            eg.ActionGroup,
            'Media',
            'Media Controls',
            'Media Controls',
            (
                (
                    SendAction,
                    "fnMuteToggle",
                    "Toggle mute",
                    "Mute volume.",
                    "q"
                ),
                (
                    SendAction,
                    "fnPauseToggle",
                    "Toggle Play/Pause",
                    "Play/pause media playback.",
                    "{Space}"
                )
            )
        ),
    )
    
    # The above does not need to be spread out like it is.
    # I did that so it would be easier to understand the comments above it
    # You can format it like so.
    
    
    ACTION_LIST = (
        (eg.ActionGroup, 'AudioControls', 'Audio Controls', 'Audio Controls', (
            (SendAction, "fnPause", "Pause", "Pauses playback.", "{Space}"),
            (SendAction, "fnVolumeUp", "Volume Up", "Increase audio volume.", "+"),
        )),
        (eg.ActionGroup, 'MediaControls', 'Media Controls', 'Media Controls', (
            (SendAction, "fnMuteToggle", "Toggle mute", "Mute volume.", "q"),
            (SendAction, "fnPauseToggle", "Toggle Play/Pause", "Play/pause media playback.", "{Space}"),
        ))
    )
    
    # you can set the last item in the action tuple to any kind of a data
    # structure you want and you can evaluate the type of data that is present and
    # do whatever it is that is needed to accommodate sending properly.
    # you can also use a config dialog if there is more then a single device being
    # set to so the user has the ability to properly set that when they want to
    # run the action
    
    # be sure not to use any special characters, spaces or start with a number
    # for the class names. Till make the action difficult to use from a
    # python script.
    
    # setting the class names of the actions to have an fn before them makes it
    # easier to identify the class name. so if I wanted to call the action
    # from a python script it would be
    # eg.plugins.SomePlugin.fnVolumeUp()
    
    
    # to add the actions to your plugin you just have to pass the list.
    # EG has been designed to recognise this layout.
    
    def __init__(self):
        self.AddActionsFromList(ACTION_LIST)

topic2k:
Thank you for your contribution.

kdschlosser:
well you are now one of the developers. your name as been added to the list of people who have helped make EventGhost what it is today.

PR #331, opened 2018-11-19
Serial plugin: Updates and new features

GruberMarkus:
Renamed self.serial to self.SerialPortInterface for easier code reading and to avoid possible misunderstandings.

Smaller changes in code and GUI formatting.

It is now possible to pause and unpause the thread reading serial data and generating events out of it.

Shorter load time due to enumerating serial ports via registry and no longer via WX framework.

### DIFF
--- a/plugins/Serial/__init__.py
+++ b/plugins/Serial/__init__.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is a plugin for EventGhost.
-# Copyright © 2005-2016 EventGhost Project <http://www.eventghost.org/>
+# Copyright © EventGhost Project <http://www.eventghost.net/>
 #
 # EventGhost is free software: you can redistribute it and/or modify it under
 # the terms of the GNU General Public License as published by the Free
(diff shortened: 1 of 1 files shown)


kdschlosser:
you have to be careful when changing any method names or the parameter keywords or orientation of the parameters. There is a strong possibility of breaking API for other plugins.

I write some code that will allow for C/C++ style function overloading for methods in python. this will provide a nice mechanism to be able to shift API's over time if needs be. so if you wanted to change a functions parameters. You would simply create another method with the same name and supply the new parameters. so when the method gets called it looks at the parameter types being passed and looks for a method that takes those parameter types. in this.


It is very hard to resist the want to change the names on things so they match properly to what they do. instead of how they are. The best thing to do in this case is add a doc string and document what the existing method does. this way if someone goes to look they will be able to read what it actually does. EG has a whole slew of voodoo code in it. as an example just to open a config dialog for a plugin the code to do that is strewn across 20+ files and is subclass of a subclass of a subclass of a subclass of a class. kind of a thing. It's a pain to try and follow the data path through anything in EG.  another HUGE reason not to change the name of anything. trying to figure out where a problem is stemming form can be a nightmare.

plugins/Serial/__init__.py
```diff
@@ -111,14 +72,42 @@ def MyHexDecoder(input):
 ]
 
 
+def RevokeWorkPermit(self):
+    if self.plugin.ReceiveThread is not None:
+        if self.plugin.ReceiveThread_WorkPermit.is_set() is True:
+            self.plugin.ReceiveThread_WorkPermit.clear()
+            time.sleep(self.plugin.SerialPortInterface.timeout)
```
kdschlosser: the use of time.sleep should never be used to stall a thread loop. The reason why is there is no exit in the event EG gets shutdown. the thread will simply get stopped and the file for the port will not get closed properly. Can cause issues.

you already have a threading.Event item right there use that instead of the time.sleep. a threading.Event() instance will bump out of the wait if the event gets set.

    if self.plugin.ReceiveThread is not None:
        if self.plugin.ReceiveThread_WorkPermit.is_set():
            self.plugin.ReceiveThread_WorkPermit.clear()
            self.plugin.ReceiveThread_WorkPermit.wait(self.plugin.SerialPortInterface.timeout)

plugins/Serial/__init__.py
```diff
@@ -111,14 +72,42 @@ def MyHexDecoder(input):
...
         eg.RawReceiverPlugin.__init__(self)
-        self.AddAction(Write)
-        self.AddAction(Read)
-        self.serial = None
+        self.AddAction(fnWrite)
+        self.AddAction(fnRead)
+        self.AddAction(fnPauseEventGeneration)
+        self.AddAction(fnFlushInput)
+        self.AddAction(fnContinueEventGeneration)
+        self.SerialPortInterface = None
```
kdschlosser: the use of the fn in the last example we typically only use when generating the actions from a list. we do this as an identifier to let us know that the class is dynamically created for the action. it is not something that is hard coded. if it is a hard coded class stick to the PEP8  camel case naming convention.

kdschlosser:
I did a quick example with this plugin. this should be how the layout of all plugins are done.
I did not test this and I know I probably missed some parts (I think the \_\_stop\_\_

if there is a thread loop it should be run in it's own class. and not looping a method in the plugin it's self.
because of the use of a threadloop you have to make the plugin thread safe. what that means is that if by chance a user calls plugin.Write() from one thread and plugin.Read() from another you don't want them to end up knocking heads.

Therre is also no reason to start and stop the loop when writing or reading. let the loop do it's thing. we use thread.Event() instances coupled with a read buffer and a write buffer to handle the rest. let the loop fill the buffers like it should. less overhead then stalling the loop to do a specific read task. which is what the loop is designed to do anyway.

I also moved all of the wx.widget control labels into a single class that inherits from eg.TranslateableStrings. this provides internationalism support. there is a little mystical code magic at work with the text class. all you need do is set the class that holds all the text into plugin class at the class level. do not construct it. we do not need an instance we need the class to be in there. the text class will automatically be propagated to all actions of the plugin as self.text and that propigation will point to the matching class name for that action that is in the main Text class. You will understand more when you see it.

I also removed the use of pywin32 from this plugin. pywin32 was great back in the day when EG was made. the problem with it is it's not finished for one. and there are bugs in it that don't seem to get fixed or take a really long time to get fixed. all of the bits that are used in the Serial Plugin can be directly accessed via ctypes. which is what I did. It is better to use as much of the stdlib as you possibly can. because when packages get updates it breaks all kinds of things.. It's less work in the long run.

attached is the plugin. use it as an example.
you do not have to do any of the code that is in it. or if you want to give it your own "style" and verbage. but try to keep to the same form
[serial_plugin.zip](https://github.com/EventGhost/EventGhost/files/2596842/serial_plugin.zip)

plugins/Serial/__init__.py
GruberMarkus: This sleep is not for stalling the thread, but to wait until the current loop of the thread is processed to avoid errors (the threading.Event() only blocks a thread at certain points in the code, where "wait" is defined).

GruberMarkus:
Thanks for your feedback! I really appreciate it, as this is just my second Python script and I usually do no programming at all. I will have a look at your sample code; due to time constraints it will probably take me a week or two until I can provide an update (if I understand the more sophisticated code at all).

The reason why I have updated the code of the original Serial plugin is that the Read function in the original version can not work while the option "generate events" is enabled.

Although this is mentioned in the plugin's documentation, I was looking for a way to safely capture feedback from writer operations without having to deactivate event generation and activate it again, each manually. Threading.Event() seems to be the solution to me, as it allows to programmatically pause and unpause the event generation thread, so that the read thread can safely access the input buffer.

plugins/Serial/__init__.py
kdschlosser: I do not remember off hand how long EG waits for threads to terminate I would have to go and check. But I do know that if there is no exit strategy from a thread being temporarily stalled EG will gripe about it in the debugging log. The main reason is because of things not getting closed properly and EG simply just shutting down, It can lead to orphaned sockets where only a reboot can fix the issue. or a lock not getting removed on a com port, and the fix is the same as the socket. a reboot would have to take place. It's just good practice to always provide a mechanism that will exit a pause in a thread.

here is an example. read the commented lines it will tell you how to use the example. you will understand what I am talking about.

    import threading
    
    event = threading.Event()
    
    def do():
        print 'thread starting'
        while not event.isSet():
            print 'thread looping'
    
            # run the script once.
            # comment and uncomment the lines below as labeled
            # run the script again
    
            # comment me
            time.sleep(10.0)
    
            # uncomment me
            # event.wait(10.0)
    
        print 'thread exiting'
    
    t = threading.Thread(target=do)
    t.start()
    
    import time
    event.wait(1.0)
    
    print 'setting event'
    event.set()
    print 'script should exit now'
    
    start = time.time()
    t_count = 0
    
    while t.isAlive():
        if t_count == 0:
            if time.time() - start >= 1:
                t_count += 1
                print 'script waiting to exit:', t_count, 'seconds'
        if time.time() - start - t_count >= 1:
            t_count +=1
            print 'script waiting to exit:', t_count, 'seconds'
    
    print 'script finally exiting'

topic2k:
My 2 cents: Make this a new plugin (SerialThreaded).
Maybe add a note to the description of the original plugins description to guide to the threaded version (with pro/cons?).

GruberMarkus:
I think this is a great idea. I will withdraw this pull request, but will take into consideration all the comments and samples posted here.

As soon as I think the code is worth it, I will create a new pull request for a new plugin.

Thanks for your support!

Issue #338, opened 2018-11-28
Exception: Timeout while calling OnComputerResume


GruberMarkus:
I am playing around with serial port interfaces and come across a reproducable problem that happens when a computer resumes from sleep.

I took the UIR plugin and modified it by adding OnComputerSuspend and OnComputerResume functions (see attachment).

Both function seem to be respected, although no triggers or print statements appear in the log.

As soon as the modified plugin is added, a whole bunch of error messages appears in the log after resume:

13:25:38   ---> Welcome to EventGhost <---
13:25:42   DEBUG: 0 MainMessageReceiver: PowerBroadcastNotifier.OnPowerBroadcast(hwnd=2033918, uMsg=536L, wParam=32787L, lParam=182526736)
13:25:42   System.AwayMode.Entering
13:25:42      On Standby
13:25:42         Display Image: BlackFullScreen
13:25:42         Turn Mute On
13:25:42         Disable: Initiate Standby
13:25:42         DVBViewer: Close DVBViewer: False
13:25:42            DEBUG: 0 ActionThread: CloseDVBViewer.__call__(waitForTermination=False)
13:25:42            DEBUG: 0 ActionThread: CloseDVBViewer.__call__ => False
13:25:42         CyberLink PowerDVD: Stop
[...]

PR #339, opened 2018-11-30
[PowerDVD] switch from WindowMatcher to ctypes, PEP8 formatting

GruberMarkus:
* WindowMatcher is very slow, even with timeout specified, which may cause problems when system suspends or enters away mode. Switching to ctypes (FindWindow, FindWindowEx).
* PEP8 formatting rules applied.

### DIFF
--- a/plugins/PowerDVD/__init__.py
+++ b/plugins/PowerDVD/__init__.py
@@ -25,7 +25,7 @@ eg.RegisterPlugin(
     kind="program",
     author="Bitmonster, GruberMarkus",
     guid="{4DBDFFA7-9E47-4782-B843-B196C74DE3EF}",
-    version="2.1",
+    version="3.0",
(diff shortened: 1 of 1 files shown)


kdschlosser:
nice work on this. Only one thing I would like you to do, do a major version bump. so go to 3.0 only because that is a pretty significant change. and it also removes the module level variable that stores the finder. which i really do like.

eg.WindowMatcher has a heap of functionality. but in some cases it is way to much functionality. 

This fix does remove that issue with the timeouts in the action thread yes?

kdschlosser:
I wanted to make a note here. if you read #338 This is going to explain a long standing issue in EventGhost. one that has eluded me for a long time as to what could be causing the problem. 

It's eg.WindowMatcher. we need to do something to speed the thing up. or we need to change how Windows notifications are handled.

PR #340, opened 2018-12-01
Fixes restart & plugin install

kdschlosser:
the title states it all

### DIFF
--- a/eg/Classes/Document.py
+++ b/eg/Classes/Document.py
@@ -486,7 +486,7 @@ class Document(object):
                     self.frame = eg.mainFrame = eg.MainFrame(self)
                     self.frame.Show()
                 else:
-                    self.frame.Iconize(False)
+                    self.frame.Raise()
                 event.set()
 
             if not threading.current_thread() == eg.mainThread:
--- a/eg/Classes/MainFrame/__init__.py
+++ b/eg/Classes/MainFrame/__init__.py
@@ -581,9 +581,6 @@ class MainFrame(wx.Frame):
         else:
             wx.Frame.Iconize(self, flag)
 
-            if not flag:
-                self.Raise()
-
             for dialog in self.openDialogs:
                 dialog.Iconize(flag)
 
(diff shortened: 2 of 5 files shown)


topic2k:
When i started EG with -debug and then call Restart from menu, the debug option is set again. Don't know if this is the intended behaviour or if the option should be given to the new instance.

kdschlosser:
the -debug is not passed to the new instance. what happens is when you set the -debug parameter and then close EG that parameter gets written to the config.py file. so when the new instance loads it see this parameter in the config.py so it loads EG in debugging mode.

you can test this if you open EG with the -debug. then close it and reopen it again without the -debug set from the CLI. it should end up in debugging mode. I think.

unless something has been hard coded into the restart menu option that launches the subprocess that passes the parameters to the new session. I have never messed around with that end of things. all i have ever changed is the cli file and some minor changes to the startup.

I am unsure why you are getting that attribute error. unless you did not compile EG. That specific attribute is built into the EventGhost.exe file when it gets compiled.

if you go to _build/data/py2exebootscript.py you will see the stderr replacement. that is to stop the annoying open notepad on error that has been coded into the error. I honestly do not know why that was ever changed because the only time an error could happen that would cause that message to open up is when we are developing the core of EG. I personally do not need a popup box asking me to open the log file. I know if i am altering the core to have that file handy to check it if EG does not load.

see what happens is if there is some kind of a problem. or if i want to simply write some logging data that stupid message popup will show up and it will hand the second instance of EG until the user closes the thing. so if the user is not about for a bit. and something like 500 events get triggered using the CLI then they will be faced with low memory and possibly a system crash simply because of that popup box for opening the log.

Now me personally I am for removing that popup box. it will alleviate the possibility of having hung instances of EG. I personally would prefer it if there were no message boxes opened up because of some startup thing going on. 

as an example the restart. that shouldn't open anything up either. so if the -restart is used from the CLI and not the menu then we can have it trigger an event instead that the user can act on. otherwise the menu option inside of EG should handle all of the checking for saved data.

kdschlosser:
yeah I saw that flash as well. I forgot to look into why it was doing it.

kdschlosser:
Oh I know hwy i added that. to bring the Window to the front. otherwise it will not bring it to the front.
slight code change needed to fix that issue

kdschlosser:
I wanted to drop a comment just to make sure you know that I have added some commits to this PR

PR #341, opened 2018-12-16
[VLC] add 64 bit support, fix play, volup and voldown

GruberMarkus:
- Add support for 64 bit VLC install directory detection via a registry query. 
- Switch to socket.sendall, as pure sendall is now deprecated.
- "volup" and "voldown" are no longer supported by VLC as standalone commands, they now require a parameter with the number of steps by which to increment or decrement the volume. E.g., "volup 1" to increase volume by one step.
- The play command now does what it says, that is start media playback. It no longer toggles between play and pause states, as there is a separate command for this.

### DIFF
--- a/plugins/VLC/__init__.py
+++ b/plugins/VLC/__init__.py
@@ -82,22 +82,18 @@ def GetVlcPath():
     """
     Tries to get the path of VLC's executable through querying the Windows
     registry.
+    Not querying HKLM\Software\VideoLAN\VLC to avoid multiple queries
+    neccessary due to registry virtualization on 64 bit systems.
     """
     try:
-        return _winreg.QueryValue(_winreg.HKEY_LOCAL_MACHINE, "SOFTWARE\Classes\VLC.OPENFolder\shell\Open\command").split("\"")[1]
-    except WindowsError:
-        if "ProgramFiles(X86)" in os.environ:  # 64 bit Windows, indepedent from Python bitness
-            if os.path.exists(os.getenv("ProgramFiles") + "\\VideoLAN\\VLC\\vlc.exe"):  # check for 64 bit VLC in default install directory
-                return os.getenv("ProgramFiles") + "\\VideoLAN\\VLC\\vlc.exe"
-            elif os.path.exists(os.getenv("ProgramFiles(X86)") + "\\VideoLAN\\VLC\\vlc.exe"):  # check for 32 bit VLC in default install directory
-                return os.getenv("ProgramFiles(X86)") + "\\VideoLAN\\VLC\\vlc.exe"
-            else:
-                return None
-        else:  # 32 bit Windows, independent from Python bitness
-            if os.path.exists(os.getenv("ProgramFiles") + "\\VideoLAN\\VLC\\vlc.exe"):  # check for 64 bit VLC in default install directory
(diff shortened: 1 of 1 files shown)


kdschlosser:
Question. Do you think maybe using the registry for the install path would be better?. Using the install path from the registry should also cover if the user set a custom install path when installing vlc

kdschlosser:
cool deal. and from the standpoint of someone just looking at the code. the registry version is somewhat easier on the eyes. 

I did want to say thank you for all of the work you have been putting in updating these plugins. it is very appreciated. some of the work you are doing is going to make my life easier. I do know that.

And you also exposed the solution to a problem that i have been chasing for some time. the thing with the timeouts happening with the power events. If i go back and look I bet every single time that issue has come up they had a find window action taking place during that event. I need to come up with a better mechanism to handle those events. we want to stall the notification system for a bit to give allow for any hardware changes to be made. but the long processing time caused by the Find Window action is not acceptable. either I am going to have to remove the stalling of the notification thread and remove that timeout or I am going to have to do something to improve the Window Matcher speed. I do not know if this is something you might be willing to tackle.

I can give you more information if you are interested.

GruberMarkus:
> either I am going to have to remove the stalling of the notification thread and remove that timeout or I am going to have to do something to improve the Window Matcher speed. I do not know if this is something you might be willing to tackle.
> 
> I can give you more information if you are interested.

I am afraid, this is beyond my limits as I am not a programmer at all.

kdschlosser:
I'm not either. I started programming a little over 3 years ago. 
Also why are you saying you lost the race before you even ran it?? 
I bet you would have no problem figuring it out. You are doing an awesome job so far. You are learning extremely fast. You have to have some kind of programming background, or language skills of some sort. Your ability to comprehend it is amazing.

GruberMarkus:
Thank you for the compliments!

I'm actually more of a script kiddie than a programmer, even though I sometimes have my bright moments. It's not a lack of interest, but much more limited time that keeps me from working even deeper into the issue.

Whenever I see an opportunity to contribute to EventGhost, I will do so - not least out of self-interest, because my HTPC setup now depends massively on EventGhost.

kdschlosser:
the best part about EG is it can handle just about anything you throw at it. This is the reason why i dug deeper into it. I saw the potential of it. how easy it is to extend. it is actually mind blowing to think where I was at 3 years ago. If i had not had someone suggest EG to me I would probably still be hunting for an "All in one" HA system. The thing about EG is it really doesn't do anything in terms of Automation. The plugins are what do it. EG provides a common place for plugins to be able to interact with one another.

plugins/VLC/__init__.py
```diff
@@ -812,14 +813,14 @@ def Configure(self, host="localhost", port=1234, showFeedbackEvents=True):
...
+        'volup 1'
     ),
     (
         ActionPrototype,
         'VolumeDown',
         'Volume Down',
         'Volume down',
-        'voldown'
+        'voldown 1'
     ),
```
topic2k: Please add a short comment about changed/fixed actions to the PR description. (And maybe replace `.)` with `-` which will be formated as list.)

GruberMarkus: Thanks for the hint, I updated the initial comment accordingly.

Merry Christmas!

PR #353, opened 2019-01-06
Fix right click on empty psace in log don't show context menu

### LINKED ISSUE #352: right click in empty space in log window doesn't open context menu (dclick needed)
Probably not related to #348, but is it intented that the right click menu takes a double click to open when i click below the log instead of a single click when i click one of the log rows?

_Originally posted by @Joyomojo in https://github.com/EventGhost/EventGhost/issues/348#issuecomment-450649972_

topic2k:
Fixes #352

### DIFF
--- a/eg/Classes/MainFrame/LogCtrl.py
+++ b/eg/Classes/MainFrame/LogCtrl.py
@@ -105,8 +105,7 @@ class LogCtrl(wx.ListCtrl):
         self.Bind(wx.EVT_MENU, self.OnCmdClearLog, id=menuId)
         self.contextMenu = menu
 
-        self.Bind(wx.EVT_RIGHT_UP, self.OnRightUp)
-        self.Bind(wx.EVT_LIST_ITEM_RIGHT_CLICK, self.OnRightUp)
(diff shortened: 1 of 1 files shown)


kdschlosser:
it looks good. my question is why did this all of a sudden break? or has this bug been since the wxPython update?

topic2k:
I guess it had been there but no one noticed it.

kdschlosser:
I just tried an older version of EG and the context menu works fine..

so now the question is what did we do that has cause this issue????...

I can't think of anything off hand that has caused this to simply just happen... OOoo 

maybe it is only happening in windows 10?

topic2k:
which version did you test?

kdschlosser:
a WIP that i have

kdschlosser:
```python
string = 'WIP-2018.05.25-17.43.19'
base = 'WIP'
major = 0
minor = 0
patch = 0
alpha = 0
beta = 0
rc = 0
buildTime = 1527270199.37
```

gemboot:
> 
> 
> I guess it had been there but no one noticed it.

yes it's there for at least one full year, I've noticed it back then but since it was bothering the usage of EG i've never took the time to report it. So I can confirm that it's not something new :)

Issue #351, opened 2019-01-06
Exepction when trying to open the options dialog on system with en_US locale, 0.5.0-rc6


thecode:
**Describe the bug**
Exepction when trying to open the options dialog on system with en_US locale
 
**To Reproduce**
Steps to reproduce the behavior, something like:
1. Go to 'File'
2. Click on 'Options'
3. Error:
Traceback (most recent call last) (0.5.0-rc6):
  File "C:\Program Files (x86)\EventGhost\eg\Classes\TaskletDialog.py", line 162, in ProcessingTask
    self.Configure(*args, **kwargs)
  File "C:\Program Files (x86)\EventGhost\eg\Classes\OptionsDialog.py", line 194, in Configure
    languageChoice.SetSelection(languageList.index(config.language))
ValueError: 'en_US' is not in list

**Expected behavior**
Options dialog should be displayed

**Additional context**
As a workaround I have changed line 84 in OptionsDialog.py from:
`        languageList = ["en_EN"]
`
to:
`        languageList = ["en_US"]
`
Options dialog opened successfully without any errors

**EventGhost version**
Software
Program Version: 0.5.0-rc6
Build Time: יום ב, 24 דצמ 2018 09:01:48
Python Version: Stackless Python 2.7.12 final 0
wxPython Version: 3.0.2.0

System
Operating System: Microsoft Windows 10 Professional, 64-bit (build 17134)

kdschlosser:
can you do me a favor.. can you fix the code and submit a PR for it please. I would like for your name to appear in the changelog and also for the history of who has helped with the project.

topic2k:
That was introduced in https://github.com/EventGhost/EventGhost/commit/9968f68ff4ffeee97404d40728b552aaf23f422e by @kdschlosser. Myabe he can remember why he changed it.

kdschlosser:
the language systems needs a little TLC. because there is an issue in wxPython (wxWidgets) that incorrectly gets the locale I had to build a cross reference table. and to do that I had to use IETF language tags. which is a combination of the iso 639-1 language code and the iso 3166-1 Alpha2 Country code. this is pretty much the standard for having a language locale id's put together.


I think that the options dialog needs to have a once over. I have a funny feeling there is something else that is missing from it.

kdschlosser:
question. did you happen to check out that language file I sent you?? I am wondering how the translation came out.

thecode:
[...]

PR #356, opened 2019-01-06
Fix issue when locale has no translation (OptionsDialog.py), and right to left bug on (SpinNumCtrl.py)

thecode:
On first load of the options dialog an auto detect language is used, if this language doesn't exists the dialog will not open with exception trying to select an invalid language from the list.
solution if to default to English language.
If later on a translation for this language is created it will be added to the language list and will not be override to English.


thecode:
Fix for issue #351

kdschlosser:
thank you for contributing We really do appreciate the help.

PR #358, opened 2019-01-18
[Builder] - Enhancement - Updates Builder StaticImports.py

kdschlosser:
Changes the name of Build StaticImport.py to Build IDE Integration (StaticImports.py)
Adds additional information into the docstring of StaticImports.py
Removes RegisterPlugin from StaticImports.py
Creates symbolic links for CorePluginModule and UserPluginModule.
Adds eg\CorePluginModule and eg\UserPluginModule to gitignore

### DIFF
--- a/_build/builder/BuildIDEIntegration.py
+++ b/_build/builder/BuildIDEIntegration.py
@@ -17,15 +17,14 @@
 # with EventGhost. If not, see <http://www.gnu.org/licenses/>.
 
 import os
+import ctypes
 from glob import glob
-from os.path import join
+from os.path import join, exists, expandvars
+from ctypes.wintypes import BOOL, DWORD
+
 
 # Local imports
 import builder
-
-SINGLETONS = (
-    "document",
-)
 
 HEADER = '''\
 # -*- coding: utf-8 -*-
@@ -47,12 +46,33 @@ HEADER = '''\
 # with EventGhost. If not, see <http://www.gnu.org/licenses/>.
 
 """
-This file was automatically created by the BuildStaticImports.py script.
+This file was automatically created by the BuildIDEIntegration.py script.
 Don't try to edit this file yourself.
 
-This module is not directly used by EventGhost. It only exists to help
-pylint and other tools to read the sources properly, as EventGhost is using
-a lazy import pattern.
+This module is used to assist and IDE in locating the modules that are apart 
+of EventGhost. Because EventGhost uses a dynamic import system and there are 
+no "hard coded" imports of it's modules, normally an IDE would not be able to 
+track these modules. This file never gets used by the program, it is only here 
+to provide the "hard coded" imports an IDE needs to see.
+
+Follow the instructions below to create an EventGhost programming environment.
+Here is a list of a couple of the features in your IDE that will now work 
+properly.
+
+    Code Completion
+    Function/Method/Class parameters
+    Code Inspection
+    Object Definitions
+
+Instructions for PyCharm
+    Build EventGhost with the Build IDE Integration selected. 
+    File --> Settings --> Project:{YOUR PROJECT NAM} --> Project Structure
+    +Add Content Root
+    Navigate to one level lower then this file location.
+    click on OK
+    click on Sources
+    click on Apply
+    click on OK
 """
 '''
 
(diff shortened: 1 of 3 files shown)


_build/builder/BuildIDEIntegration.py
```diff
@@ -47,26 +46,69 @@
...
+def symlink(source, link_name):
+    flags = 1 if os.path.isdir(source) else 0
+    lpTargetFileName = ctypes.create_unicode_buffer(source)
+    lpSymlinkFileName = ctypes.create_unicode_buffer(link_name)
+    dwFlags = DWORD(flags)
+
+    if not CreateSymbolicLinkW(lpSymlinkFileName, lpTargetFileName, dwFlags):
+        print (
+            'Build IDE Integration: '
+            'Unable to create symbolic links user MUST '
```
topic2k: Missing punctuation.

_build/builder/BuildIDEIntegration.py
```diff
@@ -47,26 +46,69 @@
...
+    Code Completion
+    Function/Method/Class parameters
+    Code Inspection
+    Object Definitions
+
+Instructions for PyCharm
+    Build EventGhost with the Build IDE Integration selected. 
+    File --> Settings --> Project:{YOUR PROJECT NAM} --> Project Structure
+    +Add Content Root
+    Navigate to one level lower then this file location.
```
topic2k: Not sure if i missunderstand of if it's wrong. My folder structure is this:
```
EventGhost-dev
    _build
        builder
            BuildIDEIntegration.py
        data
        ...
    eg
        Classes
        ...
```
From your instruction i would set `_build` as source folder, but ive set `EventGhost-dev` as sources folder. Can you clarify?

topic2k (commented):
So if i understand correct, this will only be needed once (when i setup my development environment) and not for building an EventGhost exe? Then i would propose to remove it from Builder and make it runable standalone. Show a Dialog with the information what it does and give the user the choice of continue or abort.

Do the symbolic links be visible in explorer? I ask because i think i never used symbolic links before. Can they be deleted or changed? (Mabye add that info also to the docstring.)

_build/builder/BuildIDEIntegration.py
kdschlosser: what you are reading is the header that gets placed into StaticImports.py. which when built resides in the eg folder. so one folder then the eg folder is thee main directory for eg.. this way 

    import eg

will work

_build/builder/BuildIDEIntegration.py
kdschlosser: what punctuation???

_build/builder/BuildIDEIntegration.py
```diff
@@ -47,26 +46,69 @@
...
 
 """
-This file was automatically created by the BuildStaticImports.py script.
+This file was automatically created by the BuildIDEIntegration.py script.
 Don't try to edit this file yourself.
 
-This module is not directly used by EventGhost. It only exists to help
-pylint and other tools to read the sources properly, as EventGhost is using
-a lazy import pattern.
+This module is used to assist and IDE in locating the modules that are apart
```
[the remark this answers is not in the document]
kdschlosser: https://www.grammarly.com/blog/apart-vs-a-part/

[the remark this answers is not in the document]
kdschlosser: :zipper_mouth_face:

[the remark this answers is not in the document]
topic2k: :v: sorry, i'm german :slightly_smiling_face:  i think i've got it now.

_build/builder/BuildIDEIntegration.py
topic2k: Ah, yes, i ignored that it's written to another file.

_build/builder/BuildIDEIntegration.py
topic2k: 'Unable to create symbolic links**,** user MUST '
'be in the Administrators group**.**'

kdschlosser:
Yeah I can do that. The StaticImports has always been in the builder. I simply changed the name so people will know what it actually does.

A symlink is visible in explorer. Think of a symlink like a Pointer when programming. It simply points to a location. But in reality it contains nothing it's self. You can delete a symlink like any other file or folder. Except when you delete a symlink folder it does not delete the data that is conatined in it. It simply deletes the pointer. I did the symlinks so you can do

    from eg.CorePluginModule import Webserver

And it will code complete and also show you a list of all of the plugins.

Right now there is a large chunk of code in eg/__init__.py that I can move to StaticImports. I will clean it out and do a try except to import StaticImports. If it is there then it will work it's magic. 

What I did is I have a copy of EG that is only for purposes of setting the environment. I do not modify it. It never gets run. It is only there for use with PyCharm.

PR #361, opened 2019-01-18
[Builder] - BugFix - InnoSetup.template

kdschlosser:
In newer versions of InnoSetup they dropped support for 2 languages,
These languages have been removed from the template.

There was also a line in the template that was Inc(i) It appears as
tho that command may have been in older versions of InnoSetup Unicode
In the current documentation it is not listed. So I am going to assume
it has been removed. It is the same thing as doing i := i + 1 so
it was not a big deal to change it.

### DIFF
--- a/_build/data/InnoSetup.template
+++ b/_build/data/InnoSetup.template
@@ -61,7 +61,7 @@ begin
         Sleep(500)
       end;
       Success := RenameFile(Src, Dest)
-      Inc(i)
+      i := i + 1
     until (i = 20) or (Success = True);
     if (i = 20) then begin
       Result := False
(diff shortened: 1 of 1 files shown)


_build/data/InnoSetup.template
```diff
@@ -61,7 +61,7 @@ begin
         Sleep(500)
       end;
       Success := RenameFile(Src, Dest)
-      Inc(i)
```
topic2k: [Inc()](https://www.freepascal.org/docs-html/rtl/system/inc.html) is a function of the Pascal scripting used by/for InnoSetup scripts. Did you see any error with this script?

topic2k: hmm, i can't replicate the problem :man_shrugging:

I deinstalled InnoSetup, made sure there are no .islu files left on my drives, installed InnoSetup 5.6.1 and run build script. No errors. The two .islu files exist and 'Inc' wasn't be marked as unknown identifier.

kdschlosser:
That's bonkers. I just downloaded InnoSetup and tested it. 

You are downloading the Unicode version yes?

kdschlosser:
This came up on me a while ago. And I forgot about it. I am helping a guy with setting up PyCharm and the StaticImports. And he came across the problem as well.

topic2k:
> That's bonkers. I just downloaded InnoSetup and tested it.
> 
> You are downloading the Unicode version yes?

Yes (http://files.jrsoftware.org/is/5/innosetup-5.6.1-unicode.exe)

Even if i can't reproduce the error, i can of course still merge the PR if you want.

kdschlosser:
I want to know what is so special about your setup that you are not getting the error. I hate not knowing. there is a reason and I want to know why.

topic2k:
I did a clean install on fresh virtual box. The language files exists. I guess you accidently used the non-unicode version of innosetup. There these both languages are not supported and the error with the unknown identifier `Inc` shows up.

PR #362, opened 2019-01-18
[Builder] - BugFix - CheckDependencies.py pip._veendor.requests traceback

kdschlosser:
There appears to be a glitch in version 18.1 of pip. I am not sure how
far back in the versions this problem goes. I do know that if you use
version 9.0.3 of pip when building EG everything is fine.

The problem produces a traceback when importing CheckDependencies

```python
    from pip._vendor import requests
  File "C:\Stackless27\lib\site-packages\pip\_vendor\requests\__init__.py", line 94, in <module>
    from pip._internal.utils.compat import WINDOWS
  File "C:\Stackless27\lib\site-packages\pip\_internal\__init__.py", line 40, in <module>
    from pip._internal.cli.autocompletion import autocomplete
  File "C:\Stackless27\lib\site-packages\pip\_internal\cli\autocompletion.py", line 8, in <module>
    from pip._internal.cli.main_parser import create_main_parser
  File "C:\Stackless27\lib\site-packages\pip\_internal\cli\main_parser.py", line 12, in <module>
    from pip._internal.commands import (
  File "C:\Stackless27\lib\site-packages\pip\_internal\commands\__init__.py", line 6, in <module>
    from pip._internal.commands.completion import CompletionCommand
  File "C:\Stackless27\lib\site-packages\pip\_internal\commands\completion.py", line 6, in <module>
    from pip._internal.cli.base_command import Command
  File "C:\Stackless27\lib\site-packages\pip\_internal\cli\base_command.py", line 18, in <module>
    from pip._internal.download import PipSession
  File "C:\Stackless27\lib\site-packages\pip\_internal\download.py", line 15, in <module>
    from pip._vendor import requests, six, urllib3
ImportError: cannot import name requests
```

They seem to have a broken dynamic import system. So I removed importing
requests from pip._vendor and replaced it with urllib2.urlopen

### DIFF
--- a/_build/builder/CheckDependencies.py
+++ b/_build/builder/CheckDependencies.py
@@ -24,7 +24,6 @@ import re
 import sys
 import warnings
 from os.path import basename, exists, expandvars, join
-from pip._vendor import requests
 from shutil import copy2
 from string import digits
 
@@ -429,12 +428,13 @@ def CreateVirtualEnv():
     return result
 
 def DownloadFile(url, path = "%TEMP%"):
+    import urllib2
+
     file = expandvars(join(path, basename(url.split("?")[0])))
-    r = requests.get(url, stream=True)
+    data = urllib2.urlopen(url)
+
     with open(file, "wb") as f:
-        for chunk in r.iter_content(chunk_size=16384):
-            if chunk:
-                f.write(chunk)
+        f.write(data.read())
     return file
 
 def GetChocolateyPath():
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -82,7 +82,7 @@ install:
         " "
         "--- pip 9.0.1"
         "    Updating..."
-        python -m pip install --no-cache-dir -q -U "pip==9.0.1"
+        python -m pip install --no-cache-dir -q -U pip
         "    Done."
         " "
         "--- setuptools 34.3.0"


_build/builder/CheckDependencies.py
```diff
@@ -24,7 +24,6 @@
 import sys
 import warnings
 from os.path import basename, exists, expandvars, join
-from pip._vendor import requests
```
topic2k: Didn't tested it but i guess writing just `import requests` would have been sufficent.

topic2k (changes requested):
I don't know why, but appveyor didn't recognized the changes in `.appveyor.yml` and thus used the cached environment. That means pip 9.01 were used for the build (take a look at the console output of appveyor).

In pip 18.1 the used function `get_installed_distributions()` has been moved to another sub-modul. You now need to `import pip._internal` and change the call in line 162 to `pip._internal.utils.misc.get_installed_distributions`

_build/builder/CheckDependencies.py
kdschlosser: yes but then we would have had to add requests as part of the dependencies for the build.

Now I know there is a kind of hackish way agithub is added without adding a dependency and requests would work as well. but the right way to go about it is to either use urllib2 (like i did) or add a dependency for requests

kdschlosser:
no it did invalidate the cache

Invalidated as some dependencies (.appveyor.yml) have changed

and that is also why it errored if you scroll down to the bottom. Now the question is why did it say it was successful. when it was not.

I do have to fix the get_installed_distributions. it is moved in newer versions of PIP.

PR #363, opened 2019-01-18
Updates InnoSetup version to 5.6.1

kdschlosser:
this PR needs to be merged after #361.

The AppVeyor test is going to fail. I wanted you to be able to see the error.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -105,14 +105,14 @@ install:
       $PywinPostInstall = $PythonScripts +"\pywin32_postinstall.py"
       Start-Process python -Arg "$PywinPostInstall -install -silent -quiet" -Wait
       "    Done."
-      #" "
-      #"--- Inno Setup 5.5.9"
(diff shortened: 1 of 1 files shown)


kdschlosser:
this should not have succeeded. it must be because remnants of the pre installed version on Appveyor are still kicking about....

No worries tho. you can test this yourself by uninstalling InnoSetup. make sure InnoSetup is removed from Program Files (x86) and installing the latest version.

topic2k:
Why should the build have failed? Because the build script CheckDependencies? That one only checks for minimum version, if it's above then it won't fail.

kdschlosser:
It should have failed because the InnoSetup script has not been updated. new versions of InnoSetup do not work properly with the current script. The changes to the script are in a different PR. so this was just to install th latest version. I am thinking that because there is already an InnoSetup  version installed when I install the new version it does an upgrade and leaves files in place that allow the script to run properly. But if you do a fresh clean install of InnoSetup. the build will fail.

PR #364, opened 2019-01-18
[EventGhost] - Obsolete - Removes ctypeslib as a dependency.

kdschlosser:
Removes ctypeslib as a dependency.

From what I can find ctypeslib is only used to create WinApi.Dynamic
It is no longer being maintained. In order for it to run it requires.
GCC-XML which is also no longer being maintained. It has been succeeded 
by Cast-XML which is not compatible with ctypes lib. you can download 
an old version of GCC-XML the problem is that it only works with 
MSVC 6, 7 or 7.1. There is a fork on Pypi that uses Clang, but it is 
not complete and does not function for what EG would use it for.

what ctypeslib does is it creates python versions of C/C++ code.
What EG used it for was converting some of the Windows API.

I already have 500K lines of code of the Windows API converted from 
Windows Vista to the latest build of Windows 10. So it is pretty much 
obsolete and no longer needed.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -96,7 +96,6 @@ install:
         PipInstall "py2exe 0.6.9" "py2exe_py2==0.6.9"
         PipInstall "pycrypto 2.6.1" "pycrypto==2.6.1"
         PipInstall "comtypes 1.1.3" "https://github.com/enthought/comtypes/archive/1.1.3.zip"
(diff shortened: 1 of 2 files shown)


topic2k:
As the build don't fail and a short test of the build exe didn't show any errors, i think we can merge it. But to not accidently introduce any new bugs i would merge it after the release of 0.5.

kdschlosser:
hence the reason i did not add a Milestone.

PR #366, opened 2019-01-19
Update wxPython to 4.0.4

topic2k:
This PR updates the used GUI system to version 4.0.4 (aka wxPhoenix).


topic2k:
I don't want to merge it now or in the near future, that's why i labeled it WIP. I know about the plugin problem and will continue my work to update also plugins that are not in our repo.

kdschlosser:
the wxPython thing is the single largest reason we have not upgraded to python 3. making code python 3 compatible is far easier then the wxPython changes.

So I guess with that statement. if we are going to have to update all of the plugins. should we make the move to python 3 at the same time?

topic2k:
well, personaly i would first make the change to wx4, then to python 3 and then release a beta with both changes.
The EG core and main plugins i have updatet to wx4 (1st commit to this PR) and currently i walk through plugins from the forum. Maybe next weekend i will add a first bunch of plugins to the PR.

kdschlosser:
I can get all of the plugins off the forum if you want. what I am saying is to make the plugins Python 2 and 3 compatible at the same time wee are dinking around with the wx changes. since wee are already editing the plugins it shouldn't be a huge deal to make the changes to them. I do have that script I wrote which takes care of the basic python 3 changes.

kdschlosser:
I did want to let you know. if you have not already encountered this.. wxPython 4.* is extremely unstable. I have been using it for a while now and I get app crashes quite a bit. it will be working fine and then I do a simple resize. nothing has changed other then the resize and it will crash if any plugins in EG bind to EVT_PAINT at all the whole way the DC's work is different. meaning what the rendering looks like. The plugin would have to have 2 sets of code for the 2 different wxPython versions to have it look correct.

I can replicate the app crashes and if you want some test code I can provide it. Robin has already fixed one of the problems. tho it is not yet in a release and is not anything major (I don't think it was anyway. I didn't look at how he fixed the issue). I will be sending another over to him to look at in the next few days.

kdschlosser:
I am thinking time spent could be used better then screwing around with wxPython. We should be able to write a replacement that would generate HTML/jquery code and EG could host a webpage

Most of the controls can be found already made in jquery. I know it would be a pain in the ass to do. But at least we would be moving forward instead of sideways or even a step backwards. That is kind of how I feel about wxPython. the releases are horribly slow and so are bug fixes. there are still bugs outstanding from over 10 years ago. (SpinNumCtrl) To many problems with it. and I do not want to have to monkey patch problems. and I do not feel we should use an Alpha test version either. Which is what wxPython 4.* has been for the last decade. Replacing the current version of wxPython to fix one bug and cause countless more does not seem to be the right thing to do. The users will start throwing bricks at us. LOL

It's your call to make. So whatever direction you want to take we will go that way. I hate that so much time has been spent already by you and I alike writing code for wxPython 4.*, but I would hate it even more if we both waste more and it ends up not working and causing more problems then it's worth. 

So maybe some brainstorming before we go any further with it.

PR #367, opened 2019-01-26
ConfigPanel.py: Enables Apply button on first configure

### LINKED ISSUE #357: 'OK' button is grayed in configuration popup when trying to add some plugins
**Describe the bug**
Adding plugins seems to be partially broken for me in rc5 and rc6 version as 'OK' button is not active (grayed) when adding a new plugin. I realized this only now because I added all my plugins previously in rc4. Works fine in rc4 and 0.4.1.r1722.
Some other plugins like 'Generic HID' will make the app freeze if we 'ungray' the 'OK' button by checking/unchecking an option (seems to be because we can then push 'OK' button without having selected a HID device through the plugin configuration window).

**To Reproduce**
1. Install rc5 or rc6
2. Try to add 'Keyboard' plugin (in 'Input Devices' section) for instance
3. Notice that the 'OK' button is grayed in the plugin window
4. Just uncheck the option 'Universal modifiers' in the window
5. Notice that you can now add the plugin

**Expected behavior**
'OK' button should be clickable as soon as plugin popup opens

**Configuration tested**
Windows 10 Professional (English language) build 1809

**Video**
[adding_plugins_rc5-rc6.zip](https://github.com/EventGhost/EventGhost/files/2732139/adding_plugins_rc5-rc6.zip)

kdschlosser:
If a Config dialog is opened for the first time the apply button will be enabled. On repeat opening of the dialog the Apply button will be disabled until a change is made. The act of clicking on the apply button will enable the OK button. 

The hope with this is that the user will need to actively view the settings in the config dialog and make sure they are correct before continuing. 

fixes #357

### DIFF
--- a/eg/Classes/ConfigPanel.py
+++ b/eg/Classes/ConfigPanel.py
@@ -51,7 +51,9 @@ class ConfigPanel(wx.PyPanel, eg.ControlProviderMixin):
         self.isDirty = False
         self.resultCode = None
         self.buttonsEnabled = True
-        self.dialog.buttonRow.applyButton.Enable(False)
+        self.dialog.buttonRow.applyButton.Enable(
+            self.dialog.treeItem.isFirstConfigure
+        )
         self.dialog.buttonRow.okButton.Enable(
             not self.dialog.treeItem.isFirstConfigure
         )
--- a/eg/Classes/ConfigDialog.py
+++ b/eg/Classes/ConfigDialog.py
@@ -97,6 +97,12 @@ class ConfigDialog(eg.TaskletDialog):
         self.SetAcceleratorTable(
             wx.AcceleratorTable([(wx.ACCEL_NORMAL, wx.WXK_F1, wx.ID_HELP), ])
         )
+        
+    def OnApply(self, event):
+        if self.treeItem.isFirstConfigure:
+            self.buttonRow.okButton.Enable(True)
+
+        self.DispatchEvent(event, wx.ID_APPLY)
 
     @eg.LogItWithReturn
     def Configure(self, treeItem, *args):
--- a/eg/Classes/PluginItem.py
+++ b/eg/Classes/PluginItem.py
@@ -194,7 +194,11 @@ class PluginItem(ActionItem):
             eg.actionThread.Call(self.info.Start)
 
     def SetAttributes(self, tree, itemId):
-        if self.info.lastException or self.info.initFailed:
+        if (
+            self.info is None or
+            self.info.lastException or
+            self.info.initFailed
+        ):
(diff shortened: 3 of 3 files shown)


kdschlosser:
I wanted to make it so that the usr is not going to be able to simply click on OK. I know us veteran users know to look at what is in the dialog to make changes. but there are users that o not. and will simply click on the OK button expecting everything to just work.. instead of spending some time reading. if we make the apply button available the hopes are they they will at least have top look around a bit to figure out why the OK button is not clickable.. and say to themselves.. HMMM ok i need to change this or or that..

our current issue is with plugins like the MCE Vista+ where there is no way to click the OK button to continue. at least if we allow the apply to work and when the apply button is clicked then allow the OK button after that there is a mechanism to close the dialog properly without deleting the plugin.

topic2k:
Great, now that works. Now the user is forced to press at least two buttons before he can close the dialog and hopefully stumbles upon misconfigurations. :)

kdschlosser:
this is what we were looking for initially.. I just fumbled it last time.

I had to think of the best way to go about getting to the goal. and what is going to be the best way to do it.

I think that is the cleanest code to achieve what we want done.

Do me a favor and check and make sure that it doesn't cause issues with dialogs that may not use ConfigPanel. like the Options dialog.

eg/Classes/ConfigPanel.py
```diff
@@ -51,7 +51,9 @@ def __init__(
         self.isDirty = False
         self.resultCode = None
         self.buttonsEnabled = True
-        self.dialog.buttonRow.applyButton.Enable(False)
+        self.dialog.buttonRow.applyButton.Enable(
```
topic2k: While i was testing the PR i came across plugins that don't have settings to change but still show a config dialog (File Operations for example). In that case it would be nice to have the Ok buton enabled. I found the following solution:
first you need to import
```python
from inspect import currentframe, getargvalues, getouterframes
```
and then change line 54-59 to
```python
        outer = getouterframes(currentframe())
        has_args = len(getargvalues(outer[1][0]).args) > 1
        self.dialog.buttonRow.applyButton.Enable(
            has_args and self.dialog.treeItem.isFirstConfigure
        )
        self.dialog.buttonRow.okButton.Enable(
            not has_args or
            not (has_args and self.dialog.treeItem.isFirstConfigure)
        )
```
What do you think about adding this?

kdschlosser: do it in a separate PR. if for some reason there is an issue with it then it can be easily reverted.. I do like that idea tho.

which will bring me to point out an old PR. #141

PR #368, opened 2019-01-26
Add hotkey to clear logpanel (suggested by Snowbird)

topic2k:
Adds Ctrl-L as hotkey for clearing the logpanel as suggested by Snowbird in the [forum](http://www.eventghost.net/forum/viewtopic.php?p=50326#p50326)

### DIFF
--- a/eg/Classes/MainFrame/__init__.py
+++ b/eg/Classes/MainFrame/__init__.py
@@ -335,7 +335,7 @@ class MainFrame(wx.Frame):
         Append("LogDate", kind=wx.ITEM_CHECK).Check(Config.logDate)
         Append("LogTime", kind=wx.ITEM_CHECK).Check(Config.logTime)
         menu.AppendSeparator()
-        Append("ClearLog")
+        Append("ClearLog", "\tCtrl+L")
 
         # configuration menu
         menu = wx.Menu()


topic2k:
oh, one thing, not sure if it's because of this PR:
Add a plugin that has a config dialog. In the plugin config dialog press apply and then cancel results in an exception:
```python
Traceback (most recent call last) (WIP-2019.01.26-16.28.58):
  File "wx\_core.pyc", line 16766, in <lambda>
  File "C:\Program Files (x86)\EventGhost\eg\Core.py", line 196, in Notify
    listener(value)
  File "C:\Program Files (x86)\EventGhost\eg\Utils.py", line 167, in AssertWrapper
    return func(*args, **kwargs)
  File "C:\Program Files (x86)\EventGhost\eg\Classes\MainFrame\TreeCtrl.py", line 810, in OnNodeChanged
    node.SetAttributes(self, itemId)
  File "C:\Program Files (x86)\EventGhost\eg\Classes\PluginItem.py", line 197, in SetAttributes
    if self.info.lastException or self.info.initFailed:
AttributeError: 'NoneType' object has no attribute 'lastException'
```

kdschlosser:
is the above comment in the wrong PR??

topic2k:
yes :smile:

Issue #369, opened 2019-01-28
Regional Format English (Germany) results in locale.Error: unsupported locale setting


Christoph-Wagner:
**Describe the bug**
Upon installing EventGhost, I get an error and the option to open the log, the same happens on uninstall or trying to start EventGhost. EventGhost doesn't start (but uninstalls or installs (apparently) correctly)

```
Traceback (most recent call last):
  File "EventGhost.pyw", line 53, in <module>
  File "C:\Program Files (x86)\EventGhost\eg\__init__.py", line 31, in <module>
    import Cli
  File "C:\Program Files (x86)\EventGhost\eg\Cli.py", line 34, in <module>
    locale.setlocale(locale.LC_ALL, '')
  File "locale.pyc", line 581, in setlocale
locale.Error: unsupported locale setting

```

**To Reproduce**
Steps to reproduce the behavior, something like:
1. Go to Windows Settings -> Time & Language -> Region
2. Under Regional Format, choose *English (Germany)*
3. Try to install/start EventGhost

**Expected behavior**
Allow usage.

**Additional context**
Windows 10 Pro Version 1809 Build 17763.253

I realize I have a somewhat rare setup. My keyboard layout is *US International*, Region *Germany*, Regional Format *English (Germany)* and I made additional changes to the data formats (like having an ISO date).

My workaround for now is to set the Regional Format to *English (United States)* and change all the data formats to fit with what I want. I'm not sure I can leave it like this though as I think I remember having other problems pop up with that format (my machine is used for developing apps)

**EventGhost version**
Tested with v0.5.0-rc6 and v0.4.1.r1722, error is the same but 0.4 has slightly different line numbers in the log.

Christoph-Wagner:
I'm not sure why the region matters.

Doing `locale.setlocale(locale.LC_ALL, '')` works fine and results in `'English_United States.1252'`

And `locale.localeconv()` returns my settings:

```
{
    'int_curr_symbol': 'USD',
    'currency_symbol': '€',
    'mon_decimal_point': '.',
    'mon_thousands_sep': ',',
    'mon_grouping': [3, 0],
    'positive_sign': '',
    'negative_sign': '-',
    'int_frac_digits': 2,
    'frac_digits': 2,
    'p_cs_precedes': 1,
    'p_sep_by_space': 0,
    'n_cs_precedes': 1,
    'n_sep_by_space': 0,
    'p_sign_posn': 3,
    'n_sign_posn': 0,
[...]

PR #370, opened 2019-01-29
workaround for unusual region settings

topic2k:
#369

### DIFF
--- a/eg/Classes/App.py
+++ b/eg/Classes/App.py
@@ -47,10 +47,8 @@ class App(wx.App):
     def __init__(self):
         self.onExitFuncs = []
         wx.App.__init__(self, 0)
-        lang_id = LCID_TO_WX.get(kernel32.GetUserDefaultUILanguage(), None)
 
-        if lang_id is not None:
-            self.locale = wx.Locale(lang_id)
+        self.locale = wx.Locale(eg.Cli.lang_id)
 
         self.shouldVeto = False
         self.firstQuery = True
--- a/eg/Cli.py
+++ b/eg/Cli.py
@@ -24,14 +24,24 @@ import ctypes
 import locale
 import os
 import sys
-import threading
-from os.path import abspath, dirname, join
+from os.path import abspath, join
 
-import PythonPaths
-import LoopbackSocket
+import wx
+
+from . import LoopbackSocket
+from . import PythonPaths
+from .Classes.Translation import LCID_TO_WX
+
+kernel32 = ctypes.windll.kernel32
+
+lang_id = LCID_TO_WX.get(kernel32.GetUserDefaultUILanguage(), 0)
+wx_loc = wx.Locale(lang_id, wx.LOCALE_DONT_LOAD_DEFAULT)
+li = wx_loc.FindLanguageInfo(locale.getdefaultlocale()[0].split('_')[0])
+locale.setlocale(locale.LC_ALL, li.GetLocaleName())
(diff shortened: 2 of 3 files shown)


kdschlosser:
as it turns out this problem is not a problem in python. it is a problem with Windows

build 1809 specifically. the most recent update. Microsoft sticks it to us once again.

http://www.eventghost.net/forum/viewtopic.php?f=2&t=10288&p=52829#p52829

kdschlosser:
I think that I may have located the issue.


here is the code from the locale module for setting the locale.
watch the data path specifically the locale argument

```python
def setlocale(category, locale=None):
    if locale and not isinstance(locale, (_str, _unicode)):
        # convert to string
        locale = normalize(_build_localename(locale))
    return _setlocale(category, locale)
```


we are passing "" to the local argument.
so it bypasses being normalized and gets passed as "" to the _locale extension module




ok so here is the _setlocale function in the _locale extension module.


```c
static PyObject*
PyLocale_setlocale(PyObject* self, PyObject* args)
{
    int category;
    char *locale = NULL, *result;
    PyObject *result_object;

    if (!PyArg_ParseTuple(args, "i|z:setlocale", &category, &locale))
        return NULL;

#if defined(MS_WINDOWS)
    if (category < LC_MIN || category > LC_MAX)
    {
        PyErr_SetString(Error, "invalid locale category");
        return NULL;
    }
#endif

    if (locale) {
        /* set locale */
        result = setlocale(category, locale);
        if (!result) {
            /* operation failed, no setting was changed */
            PyErr_SetString(Error, "unsupported locale setting");
            return NULL;
        }
        result_object = PyString_FromString(result);
        if (!result_object)
            return NULL;
        /* record changes to LC_CTYPE */
        if (category == LC_CTYPE || category == LC_ALL)
            fixup_ulcase();
        /* things that got wrong up to here are ignored */
        PyErr_Clear();
    } else {
        /* get locale */
        result = setlocale(category, NULL);
        if (!result) {
            PyErr_SetString(Error, "locale query failed");
            return NULL;
        }
        result_object = PyString_FromString(result);
    }
    return result_object;
}

```

in the report of the problem the error
`locale.Error: unsupported locale setting`
is what is given. there is only a single place in the code above that has
this error.

```c
    if (locale) {
        /* set locale */
        result = setlocale(category, locale);
        if (!result) {
            /* operation failed, no setting was changed */
            PyErr_SetString(Error, "unsupported locale setting");
            return NULL;
        }
```

now evn tho we are passing "" to the locale parameter for some reason it is
being interpreted as a True statement. It should not be processed here.
I believe microsoft has fixed an issue with the setting of the locale in build
1806 in the past passing "" to the set locale function for windows was how it
had to be done. and now it is not I am gathering.

I think that this is thee code block we now need to have run.
```c
    } else {
        /* get locale */
        result = setlocale(category, NULL);
        if (!result) {
            PyErr_SetString(Error, "locale query failed");
            return NULL;
        }
        result_object = PyString_FromString(result);
    }
```

so maybe changing `locale.setlocale(locale.LC_ALL, '')` to `locale.setlocale(locale.LC_ALL, None)` might fix the problem.

we can do a simple `try` `except` to catch the problem

```python
import locale


try:
    locale.setlocale(locale.LC_ALL, '')
except locale.Error:
    locale.setlocale(locale.LC_ALL, None)
    
```


thee only way to know is a user is going to have to bee running build 1806 of Windows

PR #375, opened 2019-03-03
enable OK button for config dialog without options

topic2k:
If a config dialog is called with no parameters (`Configure()`) the Ok button is enabled, so the user can leave the dialog without needing to first click the apply button.

### DIFF
--- a/eg/Classes/ConfigPanel.py
+++ b/eg/Classes/ConfigPanel.py
@@ -18,6 +18,7 @@
 
 import types
 import wx
+from inspect import currentframe, getargvalues, getouterframes
 
 # Local imports
 import eg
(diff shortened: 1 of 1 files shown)


kdschlosser:
question on this. I have not tested it yet. The way the current setup works is it checks to see if the Configure method was overridden by the subclass of eg.PluginBase or eg.ActionBase

So by default no parameters are provided to the call of Configure if the object was just created. The whole purpose to to graying out the OK button is to make sure the user double checks the settings beforee being able to click on the OK button. this would bypass that being done. I do not know if that is the behavior we want to take place.

Now I know that in the past I have used the Configure dialog to give more detailed information on what an Action does or what a plugin does and there are no parameters to be used. In these cases enabling the OK button would be warranted. But in the cases where there is some kind of a control added then we would want to keep that button grayed out.

there are 2 ways we can handle this. the first way would be to do a nested iteration over the panel.sizer to check for a widget that is not a wx.StaticText widget. if another wxSizer instance is found then that would need to be iterated over. this is a bit more complex to do.

the other is a little more straight forward, we can use inspect to check and see if there are any parameters to the Configure method much like what is done in eg.Utils.LogIt and eg.Utils.LogItWithReturn 
what we can do is a simple instance check. and if the Configure method has any parameters then enable the apply button and gray out the OK button, if there are no parameters then enable the OK button.

kdschlosser:
never mind.. HOLY COW!! i just woke up.. LOL
I was working off the description.I just looked at the code because

on first creation of the object Configure() is what gets called.

topic2k:
> the other is a little more straight forward, we can use inspect to check and see if there are any parameters to the Configure method much like what is done in 

this is what i've done.

kdschlosser:
yeah II know i saw that.. I think i was brain dead. or maybe it was your post

> If a config dialog is called with no parameters (Configure()) the Ok button is enabled, so the user can leave the dialog without needing to first click the apply button.

calling with no parameters could be interpreted as what it states literally.

```python
def Configure(self, param1='PARAM1', param2='PARAM2'):
    pass

Configure() # <----- calling with no parameters
``` 

It is my fault I should have looked at the code before I commented. But i still wanted to bust your ass a little.. :smile:

PR #376, opened 2019-03-08
[EventGhost] - BugFix - Locale permanent solution

kdschlosser:
I think you are gong to like this one @topic2k

I believe this to be the best possible way of handling the problems with `locale.setlocale`
A "locale" is define as a combination of language and country. so what I have done is this
I created a series of language and country classes. the country classes have instances of languages used in that country.
There are a total of 254 countries and 270 languages. which ends up making  a total of 610 locales
These 610 locals represent every language that Windows supports from XP to 10 including the expansion packs. Not all locals are supported by the various versions of Windows. My Windows 7 box supports 123 countries and 123 languages making up a total of 196 locales.

the downfall is wxPython which has support for 172 of the 610 locals and with my Windows 7 box there are 127 locals that are supported by wxPython

So what I have done is I have leveraged the Windows API to provide me with some needed information. I no longer set the process locale to the default user locale. I set it to the locale that is in eg.config. How am i doing this you ask??... well i found out that the string that needs to be passed to setlocale is formatted like so

    language English name_country english name.codepage

and those 3 things I am able to get from the Windows API using the iso code.

I also added more flexibility for the user. because we are no longer using the locale from windows (only for a fallback) I have split the language choice control into 2. one for the language and one for the locale. because a language can be spoken in more then a single country it would be beneficial to the user to provide them with all of the possible choices.

so i grab the language from the language files in EG. I then enumerate very country that language is spoken in and provide the user with all of the languages that EG supports along with all of the countries that the languages are spoken in. Now this is the cool part. thee languages and countries that populate the control and ones that the version of Windows they are using supports. because i did it like this the user is not going to be able to set the locale to something that is going to cause the setlocale to have a spazm.

II also provide flags for every single country. and i pull the native language and country names using the Windows API


kdschlosser:
I know I have to fix the merge conflicts.

topic2k:
I've checked out your branch and run it in my IDE. When i open the options dialog i get this error:
```python
Traceback (most recent call last) (WIP):
  File "E:\EventGhost-kdschlosser\eg\Classes\TaskletDialog.py", line 162, in ProcessingTask
  File "E:\EventGhost-kdschlosser\eg\Utils.py", line 455, in LogItWithReturnWrapper
    result = func(*args, **kwargs)
  File "E:\EventGhost-kdschlosser\eg\Classes\OptionsDialog.py", line 112, in Configure
    languages.sort(key=lambda x: x[0].english_name)
  File "E:\EventGhost-kdschlosser\eg\Classes\OptionsDialog.py", line 112, in <lambda>
    languages.sort(key=lambda x: x[0].english_name)
IndexError: list index out of range
```

kdschlosser:
OK I will have a look at it.

PR #378, opened 2019-04-14
use localized group name to determine administrator status

topic2k:


kdschlosser:
I have not tore down the code yet to see exactly what does what.

but from first look there may be one potential issue. This issue existed in original as well.

there is more then one "Administrators" Group.

DnsAdmins
Domain Admins
Enterprise Admins
Enterprise Key Admins
Hyper-V Administrators
Key Admins
Schema Admins
Storage Replica Administrators
Administrators

Some of these are local as well as Active Directory (Domain) based.

I am not sure which groups have automatic addition of other groups. And if it lists those groups when querying. I cannot remember for the life of me. I know that an Enterprise Administrator is the top one. whether or not the Administrators group is automatically I am not sure.

GruberMarkus:
There is one safe way to determine the name of the local administrators group on every Windows version supporting security groups, no matter which language the system is using: Don't query the name, but the security identifier.

The local administrators group always has the SID S-1-5-32-544.

The code seems to do exactly this and works fine here; I can not test it in a domain environment, but I think it can not differentiate between local admins and domain admins.
To be on the safe side, I would query for SID S-1-5-32-544.

There is one bug for sure, though: The script determines the general admin privileges of the account, not taking into consideration the real permissions the current process has (integrity level, admin permissions). In other words: When your account is member of the local admin group, the script will always output these permissions, but the process executing the code (eventghost.exe, for example), may have been started without admin permissions.



A user that is not a member of this group can not have administrator priviledes. Be sure to not only evaluate direct group membership, but also indirect group membership (user is in group A, group A is member of group B, group B is member of the local adminatrator group - for example, the domain group "Domain Admins" with the SID S-1-5-21<domain specific>-512 is member of the local administrator group with the SID S-1-5-32-544 on every domain joined workstation per default).

See https://support.microsoft.com/en-us/help/243330/well-known-security-identifiers-in-windows-operating-systems for details on well-known SIDs on Windows systems.

kdschlosser:
SID: S-1-5-21-519
Name: Enterprise Admins (domain)

SID: S-1-5-21-512
Name: Domain Admins (domain)

SID: S-1-5-32-544
Name: Administrators (local)

the only thing is I would have to double check on this. Just because a user is logged into a machine and may have the domain admin SID does not mean that the user is logged into the domain. and if they are not authenticated against a DC then technically speaking they do not have the domain admin rights.

The security portions of windows is a bloody mess i will say that.

GruberMarkus:
> 
> 
> SID: S-1-5-21-519
> Name: Enterprise Admins (domain)
> 
> SID: S-1-5-21-512
> Name: Domain Admins (domain)
> 
> SID: S-1-5-32-544
> Name: Administrators (local)
> 
> the only thing is I would have to double check on this. Just because a user is logged into a machine and may have the domain admin SID does not mean that the user is logged into the domain. and if they are not authenticated against a DC then technically speaking they do not have the domain admin rights.
> 
> The security portions of windows is a bloody mess i will say that.

The SID of the local administrators group is correct, the other SIDs do not exist. They miss the dynamic part identifying an Active Directory domain and the root domain identifying a forest.

https://support.microsoft.com/en-us/help/243330/well-known-security-identifiers-in-windows-operating-systems has details on how this works.

kdschlosser:
OK i see what you are saying.. the "domain" in the SID is the dynamic bit. and that can be easily gotten.

I still have to have a look to see if the domain SID still shows up even if the server never authenticated. The machine I am on right now is like that. It still allows the logon but uses some form of a cached authentication.

kdschlosser:
I am correct. The groups SIDs do show up even if I am not properly authenticated with a DC

```
Group Name                           Type             SID                                            Attributes
==================================== ================ ============================================== ===============================================================
Everyone                             Well-known group S-1-1-0                                        Mandatory group, Enabled by default, Enabled group
*********\HelpLibraryUpdaters        Alias            S-1-5-21-**********-**********-**********-1007 Mandatory group, Enabled by default, Enabled group
BUILTIN\Performance Log Users        Alias            S-1-5-32-559                                   Mandatory group, Enabled by default, Enabled group
BUILTIN\Users                        Alias            S-1-5-32-545                                   Mandatory group, Enabled by default, Enabled group
BUILTIN\Administrators               Alias            S-1-5-32-544                                   Mandatory group, Enabled by default, Enabled group, Group owner
NT AUTHORITY\INTERACTIVE             Well-known group S-1-5-4                                        Mandatory group, Enabled by default, Enabled group
CONSOLE LOGON                        Well-known group S-1-2-1                                        Mandatory group, Enabled by default, Enabled group
NT AUTHORITY\Authenticated Users     Well-known group S-1-5-11                                       Mandatory group, Enabled by default, Enabled group
NT AUTHORITY\This Organization       Well-known group S-1-5-15                                       Mandatory group, Enabled by default, Enabled group
LOCAL                                Well-known group S-1-2-0                                        Mandatory group, Enabled by default, Enabled group
                                     Unknown SID type S-1-5-21-**********-**********-**********-512  Mandatory group, Enabled by default, Enabled group
                                     Unknown SID type S-1-5-21-**********-**********-**********-1108 Mandatory group, Enabled by default, Enabled group
                                     Unknown SID type S-1-5-21-**********-**********-**********-520  Mandatory group, Enabled by default, Enabled group
                                     Unknown SID type S-1-5-21-**********-**********-**********-518  Mandatory group, Enabled by default, Enabled group
                                     Unknown SID type S-1-5-21-**********-**********-**********-519  Mandatory group, Enabled by default, Enabled group
                                     Unknown SID type S-1-18-1                                       Mandatory group, Enabled by default, Enabled group
                                     Unknown SID type S-1-5-21-**********-**********-**********-572  Mandatory group, Enabled by default, Enabled group
Mandatory Label\High Mandatory Level Label            S-1-16-12288                                   Mandatory group, Enabled by default, Enabled group, Local Group
```

specifically these ones
```
                                     Unknown SID type S-1-5-21-**********-**********-**********-512  Mandatory group, Enabled by default, Enabled group
                                     Unknown SID type S-1-5-21-**********-**********-**********-1108 Mandatory group, Enabled by default, Enabled group
                                     Unknown SID type S-1-5-21-**********-**********-**********-520  Mandatory group, Enabled by default, Enabled group
                                     Unknown SID type S-1-5-21-**********-**********-**********-518  Mandatory group, Enabled by default, Enabled group
                                     Unknown SID type S-1-5-21-**********-**********-**********-519  Mandatory group, Enabled by default, Enabled group
                                     Unknown SID type S-1-18-1                                       Mandatory group, Enabled by default, Enabled group
                                     Unknown SID type S-1-5-21-**********-**********-**********-572  Mandatory group, Enabled by default, Enabled group
```

starred out the sensitive bits.

PR #400, opened 2020-04-30
Fixes traceback when checking to see if a user is an Administrator.

kdschlosser:
http://www.eventghost.net/forum/viewtopic.php?t=10258


gerbs-11:
Why has this not been merged yet?

Issue #410, opened 2020-08-12
Action "Turn monitor on" shifts cursor


maffe:
https://github.com/EventGhost/EventGhost/blob/177be516849e74970d2e13cda82244be09f277ce/plugins/System/__init__.py#L122
I see [why](https://stackoverflow.com/questions/12572441/sendmessage-sc-monitorpower-wont-turn-monitor-on-when-running-windows-8) this is done, but please move my cursor back :-) I had this action triggered multiple times and it moved the cursor away from the position I remembered, the following blind click did something unintended :(

kdschlosser:
fair enough. Never thought that the mouse moving a single pixel would cause a problem. But here it is.. I am sorry for any issues you had because of this. You are more then welcome to make the changes needed and submit a pull request for the changes made. I would appreciate you doing that. I like to keep a record of the prople who have spent the time to locate a big and find a solution.

PR #411, opened 2020-08-13
Move cursor back to avoid accumulated movement

maffe:
Fixes https://github.com/EventGhost/EventGhost/issues/410

### DIFF
--- a/plugins/System/__init__.py
+++ b/plugins/System/__init__.py
@@ -120,6 +120,7 @@ MONITOR_STATES = dict(
 def MonitorState(state):
     if state == 'ON':
         ctypes.windll.user32.mouse_event(MOUSEEVENTF_MOVE, 0, 1, 0, 0)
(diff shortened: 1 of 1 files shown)


maffe:
I’ve seen the `sleep` call in proposed solutions, but I don’t want to introduce arbitrary delays, in my tests it worked without it.

kdschlosser:
it's a 40 millisecond delay. I am fairly sure it is there so that the Windows notification system is able to make it's callbacks before the next move takes place. I know that If you use the Move Mouse action and you do not put a wait and then add the action to click a mouse button, it doesn't work properly but if you add a wait in between the 2 everything works.

If it was a semi long delay I would be concerned with it but being that the delay is only 40 milliseconds I am not concerned with it.

kdschlosser:
also you need to push the commit with the changes for the delay in order for them to be added to the PR

